PR #6, opened 2017-12-04
Include missing "using namespace std"

spakin:
`blas_utils.hpp` was using `cout` and `endl` without reference to the C++ namespace that defines them (`std`).  This caused the C++ compiler (I'm using GCC 7.2.0 on Ubuntu 17.10) to abort with a bunch of error messages.


Ravenwater:
Thank you for the catch.

PR #7, opened 2017-12-05
Fix compile error due to missing header includes

xman:
Sample error:
universal.git/posit/../bitset/bitset_helpers.hpp:13:8: error: ‘bitset’ in namespace ‘std’ does not name a template type


Ravenwater:
Thank you

PR #28, opened 2018-02-26
Fix stringstream incomplete type error

lvandam:
```
/universal/posit/../bitblock/bitblock.hpp: In function 'std::__cxx11::string sw::unum::to_binary(sw::unum::bitblock<nbits>)':
/universal/posit/../bitblock/bitblock.hpp:572:22: error: 'ss' has incomplete type
    std::stringstream ss;
```


Ravenwater:
Thank you very much for the PR.

PR #73, opened 2019-03-26
Fixed a compiler error on mac, 

cjdelisle:
not sure why but it doesn't like it when you pass the bitblock without putting it on the stack first


cjdelisle:
Oops, that should have been another PR, anyway I just fixed compiler warnings when compiling a trivial program that includes posit_c_api.h.
Compiles with
* `clang -o xxx ../xxx.c -std=c99 -Wall -Wextra -pedantic -c`
* `gcc-7 -o xxx ../xxx.c -std=c99 -Wall -Wextra -pedantic -c`

PR #75, opened 2019-03-30
Proposal for a different approach to the C API

cjdelisle:
I wanted to open this up to start a conversation about what we might want to do with the C API.

Changes I made are as follows:

* Fix compiler warning with -std=c++14 which is an error on clang
* Change from `posit_<op><n>` to `posit<n>_<op>` which the reasoning will become apparent.
* `posit_format128(pc, str)` becomes `posit128_str(str, pc)` output first, keeping with strcpy() memcpy() etc
  * change POSIT_FORMAT128_SIZE to posit128_str_SIZE which incorporates the exact name of the function ( example of this pattern used in a popular crypto library: https://github.com/jedisct1/libsodium/blob/master/src/libsodium/include/sodium/crypto_aead_chacha20poly1305.h#L16 )
* change the types of `posit<n>_t` from int types to structures so that casually adding them to integers will be a compile time error.
  * as a result of this, added `posit<n>_bits()` which returns the bit representation as an unsigned int
* New conversion functions `posit<n>_to<ld|d|f|sll|sl|si|ull|ul|ui>()` for converting a posit to a non-posit value. Also `posit<n>_from<ld|d|f|sll|sl|si|ull|ul|ui>()` which converts from a non-posit to a posit.
* Conversions between posit types using `posit32_top16()` or `posit16_fromp32()`
* Arithmetic functions which accept posit and non-posit values and convert non-posits to posits first:
  * `posit32_t posit32_subui(posit32_t, unsigned int)`
  * `posit32_t posit32_uisub(unsigned int, posit32_t)`
* Arithmetic functions on posits of different width (in this example, the 16 is converted to 32):
  * `posit32_t posit32_subp16(posit32_t, posit16_t)`
  * `posit32_t posit32_p16sub(posit16_t, posit32_t)`
* Renamed bit_assign to `posit<n>_reinterpret()` which uses the name used in webassembly for bit casts and in C++ casting.

For posit16, these are the functions which are implemented in C++, the rest are sugar which rely on them:

```c
void posit16_str(char out[static posit16_str_SIZE], posit16_t p);

posit16_t posit16_addp16(posit16_t x, posit16_t y);
posit16_t posit16_subp16(posit16_t x, posit16_t y);
posit16_t posit16_mulp16(posit16_t x, posit16_t y);
posit16_t posit16_divp16(posit16_t x, posit16_t y);

posit16_t posit16_fromp8(posit8_t p);
posit16_t posit16_fromp32(posit32_t p);
posit16_t posit16_fromp64(posit64_t p);
posit16_t posit16_fromp128(posit128_t p);

posit16_t posit16_fromld(long double p);
posit16_t posit16_fromd(double p);
posit16_t posit16_fromf(float p);
posit16_t posit16_fromsll(long long p);
posit16_t posit16_fromsl(long p);
posit16_t posit16_fromsi(int p);
posit16_t posit16_fromull(unsigned long long p);
posit16_t posit16_fromul(unsigned long p);
posit16_t posit16_fromui(unsigned int p);

long double posit16_told(posit16_t p);
double posit16_tod(posit16_t p);
float posit16_tof(posit16_t p);
long long posit16_tosll(posit16_t p);
long posit16_tosl(posit16_t p);
int posit16_tosi(posit16_t p);
unsigned long long posit16_toull(posit16_t p);
unsigned long posit16_toul(posit16_t p);
unsigned int posit16_toui(posit16_t p);
```

Without using generics, the only preprocessor defines leaked by this header are:
* posit8_str_SIZE
* posit16_str_SIZE
* posit32_str_SIZE
* posit64_str_SIZE
* posit128_str_SIZE

## Generics

If the C version is C11 or newer and POSIT_NO_GENERICS is *not* defined before including posit_c_api.h, a set of additional macros will be created:

* `posit<n>(x)` detect what type x is and select the relevant conversion function e.g. `posit16(0)` which mimics the golang/rust style type casting
* `posit_<op>(x, y)` detect the types of x and y and then add them as posits. If x is a posit then y will be converted to the posit type that x is, if x is not a posit and y is a posit then x will be converted to that posit type.
  * `posit32_t my_add(posit32_t x, posit16_t y) { return posit_add(x, y); }` adds as posit32 because it's the first parameter
  * `posit32_t another_add(int x, posit32_t y) { return posit_add(x, y); }` adds as posit32
  * `posit32_t another_add(int x, int y) { return posit_add(x, y); }` compile time error
* `posit_<op>(x)` for a posit type x, detect the type and perform an operation:
  * `posit_str()`
  * `posit_sqrt()`
  * `posit_to<ld|d|f|sll|sl|si|ull|ul|ui>()` generic convert

## Notes:
* posit_c_api.cpp line 135 is wrong
* generic posit_add() may be too dangerous
* maintainability concerns about posit_c_macros.h ?


c_api/posit/posit_c_api.cpp
```diff
@@ -45,604 +42,139 @@ void unmarshal128(sw::unum::bitblock<128>& raw, posit128_t& a) {
...
-	pa.set_raw_bits(a);
-	pb.set_raw_bits(b);
-	presult = pa / pb;
-	return (posit16_t)presult.encoding();
-}
+	template<class ocapi>
+	static positN_t fromp(decltype(ocapi::positN) p) {
+		using namespace sw::unum;
+		posit<ocapi::nbits, ocapi::es> inp = decltype(ocapi::conv)::decode(p);
+		// TODO: There must be a better way to do this
```
cjdelisle, line 167: This is wrong, this should be using a correct posit conversion call rather than cramming everything into double and back.

cjdelisle:
Oh, btw I know why the posit128_t tests are failing (I assume it's known, they fail on c_api).

> With the GNU C Compiler, long double is 80-bit extended precision on x86 processors regardless of the physical storage used for the type (which can be either 96 or 128 bits)

https://en.wikipedia.org/wiki/Long_double

Ravenwater:
Can you rebase this to the HEAD of c_api branch?

Ravenwater:
I like this work very much. The arguments for the changes are accepted. Now we just need to marge it.

PR #77, opened 2019-03-30
C api fix posit8 test

cjdelisle:
posit8_reinterpret((float)x) effectively casts x to a float, then to a uint8_t, then reconsiders the bits of the uint8_t as the bits of the posit, so the test was failing.


cjdelisle:
argh, crap, I committed the wrong thing... just a sec

PR #78, opened 2019-03-30
C api fix posit8 test really this time

cjdelisle:
that was silly ( this is what #77 was supposed to be )


cjdelisle:
I just merged it, positN_reinterpret() is a [reinterpret_cast](https://en.cppreference.com/w/cpp/language/reinterpret_cast) which is really saying "these are the bits, I know what I'm doing" and using the low 8 bits of a double is clearly not going to work

PR #93, opened 2019-04-07
Added a test to demonstrate add_exact expected behavior

cjdelisle:


cjdelisle:
You can close this PR, it's going to slow the tests down, but here is a branch which contains a demonstration of how you should get parallel addition using add_exact(), it turns out what the mutating comparator doesn't actually work with quicksort as I had thought, but this is just as good, it does a sort/add loop until there is nothing left to do.

On a GPU this would look like sort_kernel, add_kernel, sort_kernel add_kernel etc. However, I think it's possible to do better by segmenting the list of numbers and sort/adding the sub-lists and then exchanging upper and lower halves of sub-lists with neighboring threads, if the developer is really clever, it's potentially possible to even have threads go work on other number-lists when they're done with the one they're working on (they hit all zeros)...

PR #150, opened 2020-08-27
Revert "Fix Lowest Posit Function"

Ravenwater:
Reverts stillwater-sc/universal#149

### DIFF
--- a/include/universal/posit/numeric_limits.hpp
+++ b/include/universal/posit/numeric_limits.hpp
@@ -140,8 +140,8 @@ namespace std {
 			return sw::unum::maxpos<nbits, es>(pmaxpos);
 		} 
 		static constexpr Posit lowest() { // return most negative value
-			Posit pminneg;
-			return sw::unum::minneg<nbits, es>(pminneg);
+			Posit pmaxneg;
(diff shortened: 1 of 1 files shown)


Ravenwater:
turns out that numeric_limits<>::lowest() is defined as: 

static constexpr T lowest() noexcept; (since C++11)
Returns the lowest finite value representable by the numeric type T, that is, a finite value x such that there is no other finite value y where y < x. This is different from std::numeric_limits<T>::min() for floating-point types. Only meaningful for bounded types.

This is equivalent to maxneg<>() functions defined for the individual number systems.

hypercubestart:
sorry, please correct me if I am misunderstanding something.

Lowest is defined as the most negative finite number representable by a posit.
My understanding, is that the lowest  posit number has a binary representation of all 1's, which is returned by `minneg`.
`maxneg` returns a binary representation of 1[num_bits - 2 "0's"]1, which is the largest negative value of a posit.

hypercubestart:
oh nvm I am realizing that if the sign bit is set, the rest of the bits are interpreted after taking the two's complement.

thanks for clearing this up!

PR #162, opened 2020-10-09
Inverse and in-place transpose operators

Ravenwater:
Adding Gauss-Jordan matrix inverse operator.

Adding in-place transpose and free function.

Adding row_order_index and column_order_index matrix generators.


Ravenwater (commented):
passes regression tests

PR #163, opened 2020-10-14
Add Jacobi and Gauss-Seidel examples

allanleal:
This PR adds two new examples to demonstrate the use of the universal library in the context of numerical methods.


Ravenwater (approved):
Awesome!

PR #213, opened 2021-03-09
Added a rough idea for randsvd using external dependencies

rainmaker6:
Added a rough implementation using mtl for randsvd generator.


include/universal/blas/generators/randsvd.hpp
```diff
@@ -0,0 +1,65 @@
...
+                }
+                return rank;
+            }
+
+            template<typename T>
+            boost::tuple randsvd(vector<vector<T>> A) {
+                int k = min(num_cols(A), num_rows(A));
+                int n = num_cols(A), m = num_row(A);
+                //generate a gaussian random matrix of size n x k 
+                vector<vector<T>> omega(n, vector<T>(k));
```
Ravenwater: use a proper matrix, not a vector of vectors

include/universal/blas/generators/randsvd.hpp
```diff
@@ -0,0 +1,65 @@
...
+                                    A[k][p] -= A[j][p] * A[k][i];
+                            }
+                        }
+                    }
+                }
+                return rank;
+            }
+
+            template<typename T>
+            boost::tuple randsvd(vector<vector<T>> A) {
```
Ravenwater: cannot bring boost in as a dependency.

include/universal/blas/generators/randsvd.hpp
```diff
@@ -0,0 +1,65 @@
+#pragma once
+
+#include<algorithm>
+#include<vector>
+#include<universal/blas/matrix.hpp>
+#include<mtl/operation/qr.hpp>
```
Ravenwater: cannot bring MTL in as a dependency. 
if you need these algorithms, we will need to implement them in sw::universal::blas.

include/universal/blas/generators/randsvd.hpp
```diff
@@ -0,0 +1,65 @@
...
+#include<mtl/operation/svd.hpp>
+const double EPS = 1E-9;
+double k = 0.0000001;
+namespace sw {
+    namespace universal {
+        namespace blas {
+
+            template<typename T>
+            int find_rank(vector<vector<T>> A) {
```
Ravenwater: don't use vector<vector<>>, use a proper matrix

Ravenwater (changes requested):

include/universal/blas/generators/randsvd.hpp
```diff
@@ -11,54 +13,54 @@ namespace sw {
     namespace universal {
         namespace blas {
 
-            template<typename T>
-            int find_rank(vector<vector<T>> A) {
-                int n = num_rows(A);
-                int m = num_cols(A);
+            template<typename Scalar>
+            size_t find_rank(const matrix<Scalar> A) {
```
Ravenwater: this needs to be a reference, otherwise, you are copying the whole matrix onto the stack.
size_t find_rank(const matrix<Scalar>& A) {...}

include/universal/blas/generators/randsvd.hpp
```diff
@@ -11,54 +13,54 @@ namespace sw {
...
-                vector<vector<T>> D(k, vector<T>(k));
+            template<typename Scalar>
+            void qr(const matrix<Scalar> A, matrix<Scalar> Q, matrix<Scalar> R) {
+            	//R is the upper triangular matrix
+            	//Q is the orthogonal matrix
+            	
+            }
+            
+            template<typename Scalar>
+            tuple randsvd(const matrix<Scalar> &A) {
```
Ravenwater: personally, I like to put the reference symbol next to the type, not the variable, as in:
tuple randsvd(const matrix<Scalar>&      A) { ... }
As the & symbol modifies the type, it should be associated visually with the type it is modifying.

Ravenwater (changes requested):
make certain that complex arguments are always passed by reference so that there is no unnecessary copying going on.

make certain that the reference symbol is visually associated with the type it is modifying.

Ravenwater (commented):
small nit: when you are dealing with numerical values, when you need to test for zero or non-zero you should not use operator!() as it may not be defined on the numerical type.

Line 25:
if (abs(A[j][i]) > EPS && !row[j]) break  should be

if (abs(A[j][i]) > EPS && row[j] != Scalar(0)) break;

include/universal/blas/generators/randsvd.hpp
```diff
@@ -39,27 +36,80 @@ namespace sw {
                 }
                 return rank;
             }
+            template<typename Matrix, typename MatrixQ, typename MatrixR, typename Scalar>
+            void qr(const Matrix& A, MatrixQ& Q, MatrixR& R) {
+                typedef typename matrix<Scalar>::value_type     value_type;//use Scalar
```
Ravenwater: don't use typedef's in modern C++, instead use using:

using value_type = typename matrix<Scalar>::value_type;

include/universal/blas/generators/randsvd.hpp
```diff
@@ -39,27 +36,80 @@ namespace sw {
...
+                    std::tie(Q, R) = qr(VT);
+                    D *= Q;
+                    E = triu(R, 1);
+                    V = trans(R);
+                    //have to implement when upper(R)=0
+                }
+            }
+            template<typename Scalar>
+            std::tuple<matrix<Scalar>, matrix<Scalar>, matrix<Scalar>>
+            inline svd(const matrix<Scalar>& A, double tol= 10e-10) {
```
Ravenwater: the tolerance should not be hard-typed, it should be either in the same type as the matrix elements, or if you are creating a real mixed-precision algorithm, it should be its own parameterized type.

template<typename Scalar, typename Tolerance>

include/universal/blas/generators/randsvd.hpp
```diff
@@ -39,27 +36,80 @@ namespace sw {
...
+                typedef typename Collection<Matrix>::size_type    size_type;
+                size_type    ncols = num_cols(A), nrows = num_rows(A);
+                if (nrows != ncols) std::swap(row, col);
+                matrix<Scalar> ST(ncols, ncols), V(A), D(nrows, nrows);
+                svd(A, ST, V, D, tol);
+                return std::make_tuple(ST, V, D);
```
Ravenwater: I am worried that the std::make_pair and std::make_tuple are going to negatively impact performance. For the moment, just continue with returning pair(R, Q) and tuple(ST, V, D) and once you have the functionality up and running, we can do a performance experiment.

include/universal/blas/blas_l1.hpp
```diff
@@ -169,26 +169,35 @@ void strided_print(std::ostream& ostr, size_t n, Vector& x, size_t incx = 1) {
...
 
-// 1-norm of a vector
-template<typename Scalar>
-Scalar norm1(const sw::universal::blas::vector<Scalar>& v) {
-	Scalar oneNorm = 0;
-	for (auto e : v) {
-		oneNorm += abs(e);
-	}
-	return oneNorm;
+// norm's of a vector
```
Ravenwater: you will need to rebase your branch with upstream master as the norm() infrastructure is already there.

include/universal/blas/generators/randsvd.hpp
```diff
@@ -0,0 +1,159 @@
+#pragma once
+
+#include<algorithm>
+#include<vector>
+#include<universal/blas/operators.hpp>
+#include<universal/blas/solvers.hpp>
+#include<universal/blas/generators.hpp>
+#include<universal/include/universal/blas/blas_l1.hpp>
+const double EPS = 1E-9;
```
Ravenwater: why are these constants doubles? what happens when we have a higher or lower precision/dynamic range use case?

I see EPS being used on line 24 and compared against a parameterized Scalar. So EPS needs to be managed in the precision of the computation. What is the dynamic you need cover? Is it minpos saturation, or is it a bigger number? What is the EPS trying to filter?

include/universal/blas/generators/randsvd.hpp
```diff
@@ -0,0 +1,159 @@
...
+                Q=qi[0];
+                for(size_t i=1;i<ncols && i<nrows-1;++i){
+                    tmp=qi[i]*Q;
+                    Q=tmp;
+                }
+                R=Q*A;
+                Q.transpose();
+            }//completed QR-decomposition
+            template<typename Scalar>
+            std::pair<matrix<Scalar>, matrix<Scalar>>
```
Ravenwater, line 48: two full copies of a matrix. Just make a note of this and performance benchmark this on a 10kx10k matrix.

include/universal/blas/generators/randsvd.hpp
```diff
@@ -0,0 +1,159 @@
...
+                inline qr(const matrix<Scalar>& A) {
+                //R is the upper triangular matrix
+                //Q is the orthogonal matrix
+                matrix<Scalar> Q(num_rows(A), num_cols(A)), R(A);
+                qr(A, Q, R);
+                return std::make_pair(Q, R);
+            }
+            template<Scalar, Vector>
+            void householder_factors(matrix<Scalar>& A, const Vector& v){
+                Scalar n=num_cols(A);
```
Ravenwater: this is incorrect. n is of type size_t, not the element type.

include/universal/blas/generators/randsvd.hpp
```diff
@@ -0,0 +1,159 @@
...
+            std::pair<matrix<Scalar>, matrix<Scalar>>
+                inline qr(const matrix<Scalar>& A) {
+                //R is the upper triangular matrix
+                //Q is the orthogonal matrix
+                matrix<Scalar> Q(num_rows(A), num_cols(A)), R(A);
+                qr(A, Q, R);
+                return std::make_pair(Q, R);
+            }
+            template<Scalar, Vector>
+            void householder_factors(matrix<Scalar>& A, const Vector& v){
```
Ravenwater: why is there asymmetry between matrix<Scalar> and Vector? If the vector elements are of a different type, you will get conversion dependencies and for transparency it would be good to have the explicit type of the vector elements to resolve which conversion is necessary.

include/universal/blas/generators/randsvd.hpp
```diff
@@ -0,0 +1,159 @@
...
+                                for (size_t p = i + 1; p < m; ++p)
+                                    A[k][p] -= A[j][p] * A[k][i];
+                            }
+                        }
+                    }
+                }
+                return rank;
+            }            
+            template<typename Scalar>
+            inline void qr(const matrix<Scalar>& A, matrix<Scalar>& Q, matrix<Scalar>& R) {
```
Ravenwater: needs to be in its own file under solvers

include/universal/blas/generators/randsvd.hpp
```diff
@@ -0,0 +1,159 @@
...
+                Scalar n=num_cols(A);
+                for(size_t i=0;i<n;++i){
+                    for(size_t j=0;j<n;++j){
+                        A[i][j]=-2*v[i]*v[j];
+                    }
+                }
+                for(size_t i=0;i<n;++i) A[i][i]+=1;
+            }
+            template<typename Scalar, typename Tolerance>
+            inline void svd(const matrix<Scalar>& A, matrix<Scalar>& S, matrix<Scalar>& V, matrix<Scalar>& D, Tolerance tol = 10e-10) {
```
Ravenwater: needs to be in its own file under solvers.

include/universal/blas/generators/randsvd.hpp
```diff
@@ -0,0 +1,159 @@
...
+                using value_type = typename matrix::value_type;
+                using size_type = typename matrix::size_type;
+                size_type        ncols = num_cols(A), nrows = num_rows(A);
+                value_type       ref, zero = math::zero(ref), one = math::one(ref);
+                double 	     err(std::numeric_limits<double>::max()), e, f;
```
Ravenwater: why is this type not parameterized? looks like you are assigning it to maxpos in the initialization

include/universal/blas/operators.hpp
```diff
@@ -50,6 +50,28 @@ matrix<Scalar> diag(const vector<Scalar>& d) {
...
+    for (size_t i = x; i < num_rows(A); i++){
+        for (size_t j = x; j < num_cols(A); j++){
+            tmp[i][j] = A[i][j];
+        }
+    }
+    return tmp;
+}
+//get xth column
+template<typename Scalar>
+void get_col(matrix<Scalar>& A, vector<Scalar>& v, size_t x){
```
Ravenwater, line 69: this is a copy out of the matrix. get the algorithm to work first, but we need to solve this with a column or row reference view so that there is no copying.

Ravenwater (changes requested):
the norm() functions will clash with the master branch as they are already there.

There are a couple of unparameterized variables that need some more thought. We can leave them till after validation and performance testing, but they are key algorithmic components that need to be parameterized in mixed-precision libraries.

PR #222, opened 2021-03-31
Improving ODE functionality and building convergence analysis methods

jtodd440:
- Made a file to contain definitions of ode solving algorithms
- Created a function to return the error of an ode solution and write it to a csv


Ravenwater (approved):
I am going to merge it and then clean it up a bit.

PR #261, opened 2021-11-14
Fix typo in readme

FloEdelmann:
In the install instructions, `${nproc)` is misinterpreted by Bash. It should likely be `$(nproc)` like a few lines above.


Ravenwater (approved):
great catch, thank you.

PR #285, opened 2022-06-11
Make cfloat a trivial type and add static_asserts

willwray:
Change the default constructor definition to defaulted.
Remove defaulted copy, move, copy-assign and move-assign ops.
Add is_trivial static_asserts for some cfloat types.
Add an initializer to an otherwise uninitialized cfloat in table.hpp

The PyTorch integration of cfloat hit an issue that cfloat is not trivial.
This is easily changed, but is a breaking change; default initialization
becomes a no-op rather than a call to the default constructor.

  https://en.cppreference.com/w/cpp/named_req/TrivialType

cfloat had a default constructor that explicitly zero-initializes _block.
Because it is user-defined, the constructor code is run when a cfloat is
default constructed (in principle at least), guaranteeing the storage is
zeroed out. However, because user code is run, this obstructs compiler
optimizations that treat types as 'bags of bits' with no invariants.
cfloat is therefore termed a non-trivial type.

If the constructor is =default'd then the type becomes trivial.
However, now if a cfloat type is declared without initializer:

    fp64 duble; // truble! uninitialized variable

no construcor is run, no construction is done, and the guarantee of zero-
initialization is lost - the variable remains uninitialized. Compilers
warn about uninitialized variables. In fact, making this change flagged
up an uninitialized cfloat variable in table.hpp

This behavior is expected for built-in scalar types, where it is seen as
a 'zero cost' feature despite the perils (ideally there'd be an explicit
syntax for deliberately uninitialized variables).

This is the correct behavior types like cfloat and other Universal types.

### DIFF
--- a/applications/roots/rpoly_ak1.cpp
+++ b/applications/roots/rpoly_ak1.cpp
@@ -663,7 +663,7 @@ void RealIT_ak1(int* iFlag, int* NZ, double* sss, int N, double p[MDP1], int NN,
 // iFlag - flag to indicate a pair of zeros near real axis
 
 int i, j = 0, nm1 = N - 1;
-double ee, kv, mp, ms, omp, pv, s, t = 0;
+double ee, kv, mp, ms, omp{}, pv, s, t = 0;
 
 *iFlag = *NZ = 0;
 s = *sss;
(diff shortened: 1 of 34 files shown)


Ravenwater (approved):
new pattern to follow

PR #287, opened 2022-06-26
lns skeleton

Lemurian-002:
skeleton implementation of assignment and value for logarithmic number system type


Ravenwater (approved):
good to merge

PR #294, opened 2022-07-10
Silence type-punning warning

willwray:
GCC warning:
dereferencing type-punned pointer will break strict-aliasing rules

on e.g.  uint32_t f(float f) { return *(uint32_t*)&f; }

Silenced by using a union instead.


Ravenwater (approved):
yes, that is much nicer

PR #295, opened 2022-07-10
Silence uninitialized warnings

willwray:
These are all due to type_tag(x) implementations that access their
argument x. Ironically, it looks like these were modified at some
point to access their argument in order to silence warnings about
'unused formal parameter'; a fix for that should be to remove the
formal parameter name (and its superfluous use) - I'll do a PR.


Ravenwater (approved):
thanks for the fix. With the trivial constructible semantic change these issues have crept in.

PR #296, opened 2022-07-10
Drop type_tag formal arg

willwray:
The type_tag(x) overload implementations deduce their argument type:

    template <typename X> string type_tag(X x);
 or template <typename X> string type_tag(X const& x);

The formal argument x shouldn't be used; only its type X is pertinent.
The declarations should drop the name:

  template <typename X> string type_tag(X);

An alternative signature then takes no argument; convenient when there
is no variable to use so it saves declaring one uneccesarily:

  template <typename X> string type_tag();

Here the type cannot be deduced so must be provided: type_tag<X>().
The two overloads can be combined into a one with a default argument:

  template <typename X> string type_tag(X = {});

(assuming that X is default constructible).

Note that type_tag should really be a constexpr function (or a variable
template, or a type trait) because it deals only with types. However,
relying on C++ runtime typeinfo disallows any compile-time definition,
as does returning std::string (and implementing with stringstream).

If the signature is changed to return a 'constexpr string' type then
I can PR a constexpr type_tag implementation.

### DIFF
--- a/include/universal/internal/blockbinary/blockbinary.hpp
+++ b/include/universal/internal/blockbinary/blockbinary.hpp
@@ -663,9 +663,8 @@ private:
 
 // Generate a type tag for blockbinary
 template<size_t N, typename B, BinaryNumberType T>
-std::string type_tag(const blockbinary<N, B, T>& v) {
+std::string type_tag(const blockbinary<N, B, T>& = {}) {
 	std::stringstream str;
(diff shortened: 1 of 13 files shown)


Ravenwater (approved):
totally missed that opportunity to merge the two function signatures with a default initializer. Thank you for this contribution.

I am also game for the constexpr version. The use of typeid() is convenience, so not married to it.

PR #297, opened 2022-07-10
Comment out unused retval variable

willwray:
Looks like it should've been commented out in recent change
"cleaning up blocktriple regression tests to reduce output"
which commented out retval's useage.

Commenting out rather than removing because there's a TODO.

(Third time I've prepared this change, it keeps getting lost.)
(I think this is the last of the non _block related warnings.)


Ravenwater (approved):
lgtm

PR #298, opened 2022-07-11
bit_cast polyfill

willwray:
C++20 introduced std::bit_cast, with clang, gcc and msvc implementing
it with a builtin, __builtin_bit_cast(T,v), portable between compilers
and available in < c++20 language modes.

If compiler-provided constexpr-capable bit_cast is detected it's used,
otherwise this header defines a non-constexpr bit_cast, useful as a
UB-free alternative to 'type-punning', for trivially copyable types.

The BIT_CAST_CONSTEXPR preprocessor symbol is defined as constexpr if
sw::bit_cast is constexpr, otherwise it's defined as empty, allowing
clients to propagate constexpr on functions that depend on bit_cast,
and BIT_CAST_IS_CONSTEXPR symbol is defined as true or false.

The earlier BIT_CAST_SUPPORT, and related CONSTEXRESSION, symbols are
now redundant and can be deprecated.

### DIFF
--- a/include/universal/utility/bit_cast.hpp
+++ b/include/universal/utility/bit_cast.hpp
@@ -1,9 +1,101 @@
 #pragma once
-// bit_cast.hpp: C++20 <bit> compiler directive
+// bit_cast.hpp provides sw::bit_cast, a backport of C++20 std::bit_cast
 //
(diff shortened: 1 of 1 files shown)


Ravenwater (approved):
fancy

PR #300, opened 2022-07-30
V3.58: modular and saturating lns arithmetic

Ravenwater:
Adding a modular and saturating lns arithmetic type of fixed-size, but arbitrary configuration.


theo-lemurian (approved):
LGTM

PR #302, opened 2022-08-01
plu addition

jamesquinlan:
plu(A) takes a matrix and returns P, L, and U matrices.  It is **not** a void method, because it really makes less sense to create these matrices in a different file and send references for in-place.  PLU should generate these matrices.


jamesquinlan:
Consider only 6845dfe, updated with suggestions fro Codacy

Ravenwater (approved):
lgtm

Ravenwater (approved):
good first pass

PR #303, opened 2022-08-04
lns signature change

willwray:
Change lns template signature:
  <size_t nbits, size_t rbits, ArithmeticBehavior behavior, typename bt>
> <size_t nbits, size_t rbits, typename bt, auto...x>

The default case where x... vanishes is Saturating behavior.
To opt in to Wrapping behavior, add trailing Behavior::Wrapping:

  lns<nbits, rbits, bt>                       Saturating by default
  lns<nbits, rbits, bt, Behaviour::Wrapping>  Opt in to Wrapping

This has a benefit of a short symbol size in the default case.

Unfortunately the changed signature meant changing ~50 files.
Pay attention to these file changes:

arithmetic.hpp:
2-valued Behavior enum replaces 4-valued ArithmeticBehavior class.

            Behavior::Saturating, Behavior::Wrapping

'Wrapping' replaces 'Modular' (modular only applies to unsigned integer)
'Finite' and 'Infinite' Limit are removed as not pertinent distinctions.

lns_impl.hpp:
Lots (too many) changes in one go, including implementation changes
(not intended to change functionality - review carefully)

operator== now delegates directly to blockbinary operator==
(note: blockbinary operator== is not yet declared constexpr)

operator< is simplified to return a ternary expression.

Other comparison operators are slightly refactored to avoid redundant
checking of NaN cases where possible.

Most other operators are made hidden friends.
'inline' specifiers are removed or changed to 'constexpr' as appropriate.

### DIFF
--- a/applications/adaptive/double-double.cpp
+++ b/applications/adaptive/double-double.cpp
@@ -45,9 +45,9 @@ try {
 	std::streamsize precision = std::cout.precision();
 	
 	{
-		using lns = lns<16, 10, Saturating, std::uint16_t>;
+		using LNS = lns<16, 10, std::uint16_t>;
 
-		lns a, b, c;
+		LNS a{}, b{}, c{};
 		a = 0.5;
 		b = 2.0;
 		c = a * b;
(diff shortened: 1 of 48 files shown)


Ravenwater (approved):
Thank you for this refactor, it has cleaned up the public interface nicely.

PR #328, opened 2023-02-12
bug fix in condition estimate

jamesquinlan:
Condition number estimate (condest.hpp) was incorrectly calculated because transpose() is in place, thus modifying L and U. When the formula from Cline and Moler uses L^T (and U^T), but subsequently needs to use L (and U).

In particular, solve U^Tz = b for z, then L^Tx = z for x. Finally, solve LUy = x for y.


Ravenwater (approved):
LGTM

PR #330, opened 2023-03-02
Add integration for cmath functions.

yboettcher:
As of now, this does not compile for me with Scalar being any of the universal types:
```
#include <cmath>

using Scalar = ... // arbitrary universal type

int main() {
  Scalar a = 1507.2; // arbitrary number
  Scalar b = std::sqrt(a); // arbitrary function from cmath
  return 0;
}
```

It's working for posits when using the sqrt function like this though:
```
#include <cmath>

using Scalar = ... // arbitrary universal type

int main() {
  using std::sqrt;
  Scalar a = 1507.2; // arbitrary number
  Scalar b = sqrt(a); // arbitrary function from cmath
  return 0;
}
```

However, changing the way the sqrt function is called is only something one can do in ones own code and can become very hard when a library (Eigen/Bembel) would need to be modified.

Back during my thesis, I therefore added templated overloads for these functions for posits into the `std::` namespace and this PR intends to do this for all types in universal.

I stumbled over this again, because I intended to run as many number systems as possible through a set of Eigen algorithms (for https://github.com/openjournals/joss-reviews/issues/5072), but most of them turned out to fail compiling due to not being able to call `std::sqrt` or some other such function (log, abs, atan, tan come to mind).

Currently, this is a draft, because I am not sure adding stuff into std:: is a good idea (see https://stackoverflow.com/questions/320798/adding-types-to-the-std-namespace and https://en.cppreference.com/w/cpp/language/extending_std). Additionally, most types did not provide an appropriate `enable_if_`, so they are currently not implemented. For the cfloat however, these changes allowed me to compile programs which would not compile otherwise.

### DIFF
--- a/include/universal/number/areal/areal.hpp
+++ b/include/universal/number/areal/areal.hpp
@@ -43,4 +43,7 @@
 /// math functions
 #include <universal/number/areal/math_functions.hpp>
 
+// no enable_if_areal
+// #include <universal/number/areal/cmath_integration.hpp>
+
 #endif
--- a/include/universal/number/areal/cmath_integration.hpp
+++ b/include/universal/number/areal/cmath_integration.hpp
@@ -1,0 +1,33 @@
+#pragma once
+#include <string>
+
+#define make_std_fun(name) template <typename T> sw::universal::enable_if_areal<T, T> name(const T &t) { return sw::universal::name(t); }
+
+namespace std {
+    make_std_fun(sin)
+    make_std_fun(cos)
+    make_std_fun(sqrt)
+    make_std_fun(abs)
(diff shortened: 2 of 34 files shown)


theo-lemurian:
@yboettcher the failure to compile your test program is correct behavior. std::sqrt() is only defined for native types, and thus will not match an extension type. You need to call the specialized sqrt() function that resides in the `sw::universal` namespace. If you insist on calling the std::sqrt() function, you will need to use one of the explicit cast methods, for example: std::sqrt(double(myCfloat));

Ravenwater:
not a bug

PR #335, opened 2023-04-27
Fix cfloat comparison with fp16

davidmallasen:
Hello,

Since not all of the parameters are passed to the overloaded operator, when using `fp16` aka `sw::universal::cfloat<16, 5, short unsigned int, true, false, false>` an error is thrown:
~~~
/include/universal/number/cfloat/cfloat_impl.hpp:3753:118: error: no matching function for call to ‘operator==(const sw::universal::cfloat<16, 5, short unsigned int, true, false, false>&, sw::universal::cfloat<16, 5, short unsigned int, false, false, false>)’
 3753 |         return operator<(lhs, cfloat<nbits, es, bt, hasSubnormals, hasSupernormals, isSaturating>(rhs)) || operator==(lhs, cfloat<nbits, es, bt>(rhs));
~~~
This PR fixes this.

Best,
David


Ravenwater (approved):
Excellent catch.

PR #336, opened 2023-04-28
enable quad precision floating point component extraction

yboettcher:
This PR adds the extraction of floating point components for 128 bit long doubles.

Regarding the changes:
- **value.hpp**: Depending on which floating point configuration is used, the `fraction_without_hidden_bit` is either 64 bit or 128 bit in size. To allow for this distinction, the creation of the `fraction_without_hidden_bit` variable was moved into the if expression along with the call to `extract_fp_components()`.
A new condition was added to handle 16 byte long doubles with more than 64 mantissa bits. This uses the unirsals internal::uint128 type to hold the mantissa, but works the same otherwise. One difference is that the uint128 cannot be turned into a string directly, so the output for `_trace_value_conversion` is modified slightly.
- **extract_fp_components.hpp**: An assertion was added to the existing long double version to ensure it's not called when the mantissa does not fit into a 64 bit container. A new overload was added that uses universals internal::uint128 type as mantissa container. This version uses `memcpy` instead of `reinterpret_cast` because the long double could not be `reinterpret_cast`ed into the uint128. Assuming that the memory layout of uint128 is correct, this should be right.
Also, instead of masking away the upper bits, I assert that this function is only called when the long double has more than 64 mantissa bits (i.e. the only removed bits are in the upper part of the uint128) and then use two shifts to remove the upmost bits. The value used for shifting is a part I am not certain about, as the computed shift value emits a lot of compiler warnings when building on architectures, where the long double has fewer mantissa bits.
```
/mnt/RamDisk/patch/universal/./include/universal/native/nonconstexpr/extract_fp_components.hpp:57:18: warning: shift count >= width of type [-Wshift-count-overflow]
        _fraction.upper <<= shift;
                        ^   ~~~~~
/mnt/RamDisk/patch/universal/./include/universal/native/nonconstexpr/extract_fp_components.hpp:58:18: warning: shift count >= width of type [-Wshift-count-overflow]
        _fraction.upper >>= shift;
```
- **propenv.cpp**: This also directly accesses the extract_fp_components function and needed a little modification to use uint128 when appropriate. Because the fraction variable is used later, everything between its creation and final use is moved into the if expression.

### DIFF
--- a/include/universal/internal/value/value.hpp
+++ b/include/universal/internal/value/value.hpp
@@ -277,22 +277,40 @@ public:
 		case FP_NORMAL:
 			{
 				long double _fr{0};
-				unsigned long long _63b_fraction_without_hidden_bit{0};
 				int _exponent{0};
-				extract_fp_components(rhs, _sign, _exponent, _fr, _63b_fraction_without_hidden_bit);
-				_scale = _exponent - 1;
+				// fraction without hidden bit moved into the if
+
 				// how to interpret the fraction bits: TODO: this should be a static compile-time code block
-				if (sizeof(long double) == 8) {
+				if constexpr (sizeof(long double) == 8) {
 					// we are just a double and thus only have 52bits of fraction
+
+					std::uint64_t _63b_fraction_without_hidden_bit{0};
+					extract_fp_components(rhs, _sign, _exponent, _fr, _63b_fraction_without_hidden_bit);
+					_scale = _exponent - 1;
+
 					_fraction = extract_52b_fraction<fbits>(_63b_fraction_without_hidden_bit);
 					if (_trace_value_conversion) std::cout << "long double " << rhs << " sign " << _sign << " scale " << _scale << " 52b fraction 0x" << std::hex << _63b_fraction_without_hidden_bit << " _fraction b" << _fraction << std::dec << std::endl;
 
 				}
-				else if (sizeof(long double) == 16) {
+				else if constexpr (sizeof(long double) == 16 && std::numeric_limits<long double>::digits <= 64) {
 					// how to differentiate between 80bit and 128bit formats?
+
+					std::uint64_t _63b_fraction_without_hidden_bit{0};
+					extract_fp_components(rhs, _sign, _exponent, _fr, _63b_fraction_without_hidden_bit);
+					_scale = _exponent - 1;
+
 					_fraction = extract_63b_fraction<fbits>(_63b_fraction_without_hidden_bit);
 					if (_trace_value_conversion) std::cout << "long double " << rhs << " sign " << _sign << " scale " << _scale << " 63b fraction 0x" << std::hex << _63b_fraction_without_hidden_bit << " _fraction b" << _fraction << std::dec << std::endl;
 
+				} else if constexpr (sizeof(long double) == 16  && std::numeric_limits<long double>::digits <= 128) {
+					internal::uint128 _112b_fraction_without_hidden_bit{0};
+					extract_fp_components(rhs, _sign, _exponent, _fr, _112b_fraction_without_hidden_bit);
+					_scale = _exponent - 1;
+
+					_fraction = extract_long_double_fraction<fbits>(&_112b_fraction_without_hidden_bit);
+					if (_trace_value_conversion) std::cout << "long double " << rhs << " sign " << _sign << " scale " << _scale << " 112b fraction upper 0x" << std::hex << _112b_fraction_without_hidden_bit.upper << " lower 0x" << std::hex << _112b_fraction_without_hidden_bit.lower << " _fraction b" << _fraction << std::dec << std::endl;
+				} else {
+					assert(false);
 				}
 				_nrOfBits = fbits;
 			}
--- a/include/universal/native/nonconstexpr/extract_fp_components.hpp
+++ b/include/universal/native/nonconstexpr/extract_fp_components.hpp
@@ -49,7 +49,7 @@ inline void extract_fp_components(long double fp, bool& _sign, int& _exponent, l
 	_sign = fp < 0.0;
 	_fr = ::std::frexp(fp, &_exponent);
 
-	std::memcpy(&_fr, &_fraction, sizeof(internal::uint128)); // _fraction = reinterpret_cast<uint128>(_fr);
+	std::memcpy(&_fraction, &_fr, sizeof(internal::uint128)); // _fraction = reinterpret_cast<uint128>(_fr);
 
 	// we need to remove the upper bits that are not part of the mantissa. (all bits - mantissa bits - 1). -1 because the first bit is not stored
 	// we only need to do this on the upper part of the uint128, as we asserted that the mantissa has more than 64 bits.
--- a/include/universal/native/nonconstexpr/extract_fp_components.hpp
+++ b/include/universal/native/nonconstexpr/extract_fp_components.hpp
@@ -5,6 +5,7 @@
 //
 // This file is part of the universal numbers project, which is released under an MIT Open Source license.
 #include <cmath>
+#include <universal/internal/bitblock/bitblock.hpp>
 
 /*
  * The frexpf/frexp/frexpl functions have become constexpr in C++23. Universal is using the <bit> library
@@ -30,6 +31,7 @@ inline void extract_fp_components(double fp, bool& _sign, int& _exponent, double
 // this implementation assumes an 80bit extended precision representation
 // TODO: support a full quad precision long double
 inline void extract_fp_components(long double fp, bool& _sign, int& _exponent, long double& _fr, std::uint64_t& _fraction) {
+	assert(sizeof(long double) == 8 || sizeof(long double) == 16 && std::numeric_limits<long double>::digits <= 64);
 	if constexpr (sizeof(long double) == 8) { // check if (long double) is aliased to be just a double
 		_sign = fp < 0.0;
 		_fr = (long double)(::std::frexp(double(fp), &_exponent));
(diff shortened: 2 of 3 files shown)


Ravenwater (approved):
LGTM

yboettcher:
Is there a more extensive test suite I can run on the Pi, other than the normal ctest?

Ravenwater:
if you are running the full BUILD_CI that is the full regression suite across all number systems. That just completed successfully in CI, so I think we are good to merge.

I had forgotten that I had that little struct hidden away in bitblock. bitblock was the foundation of the early posit implementation, but now, all the other number systems are based on a limb-based system instead of the bit-oriented bitblock to gain some performance.

I will see if I can elevate the uint128 into its own swimlane so that it becomes more universally useful as a building block for other algorithms inside the library.

yboettcher:
Alright building with BUILD_CI now, will report back whether the tests pass on the Pi. As far as I can tell, the CI jobs don't actually test the new additions, since they only run when the long double is 128 bits, which is most likely not the case for these jobs.

Ravenwater:
That is correct, but your propenv should at least tell you if it hits your code path.

we can add a quick test case to the posit<128,4> and posit<256,5> test suites to see if your code moves the extra bits into the posit. Having a quad precision floating-point will enable all kinds of testing for even posit<32,2> and posit<64,3> as they need more mantissa bits than double precision provides to correctly round the reference values.

Ravenwater:
One quick idea, you can use cfloat<128, 15, std::uint32_t, true, false, false> to generate arbitrary bit patterns that you can load into the long double. One good test would be to do something like this:

quad a{1.0l},b{1.0l}, c{1.0l};
--a;  // move 1ULP down from 1.0
b -= a; // get the ULP value at 1.0
posit<128,4> ref(1.0),pa(a), pb(b), pc;
pc = pa + pb;
if (pc != ref) std::cout << "FAIL\n";

yboettcher:
So the posit_truncate test fails apparently:
```
202/321 Testing: posit_truncate
202/321 Test: posit_truncate
Command: "/mnt/RamDisk/universal/build/static/posit/posit_truncate"
Directory: /mnt/RamDisk/universal/build/static/posit
"posit_truncate" start time: Apr 28 16:04 CEST
Output:
----------------------------------------------------------
posit truncate function validation: results only
trunc                                                        posit<6,2>() FAIL 1 failed test cases
round                                                        posit<6,2>() FAIL 1 failed test cases
floor                                                        posit<6,2>() FAIL 1 failed test cases
ceil                                                         posit<6,2>() FAIL 1 failed test cases
posit truncate function validation: FAIL
<end of output>
Test time =   0.01 sec
----------------------------------------------------------
Test Failed.
"posit_truncate" end time: Apr 28 16:04 CEST
"posit_truncate" time elapsed: 00:00:00
----------------------------------------------------------
```

Although to me, this looks totally unrelated to my changes.
I assume that the test is this file, which never converts any float/double/long double into anything. It simply uses `p.setbits()` and then converts back to float...
https://github.com/yboettcher/universal/blob/main/static/posit/math/truncate.cpp

Ravenwater:
This is passing in CI:

```
posit truncate function validation: results only
trunc                                                        posit<6,2>() PASS
round                                                        posit<6,2>() PASS
floor                                                        posit<6,2>() PASS
ceil                                                         posit<6,2>() PASS
posit truncate function validation: PASS
```

You can trace out which test case is failing by setting a boolean in the test main:

```
int main()
try {
	using namespace sw::universal;

	std::string test_suite  = "posit truncate function validation";
	std::string test_tag    = "truncate failed: ";
	bool reportTestCases    = false;
	int nrOfFailedTestCases = 0;

	ReportTestSuiteHeader(test_suite, reportTestCases);
...
```
if you set reportTestCases to true, the test will print out which case is failing.

yboettcher:
First up: I now got your example from https://github.com/stillwater-sc/universal/pull/336#issuecomment-1527611965 to succeed. I just messed up the ordering of source and destination for memcpy.
The CI test is still failing though, so I would assume that it is unrelated to the component extraction. This concerns me a little though. As wrong as my previous implementation was, I am surprised that none of the tests actually caught this.

I actually ran into some other strange behaviour when trying out the `nextafter` function.
In a 128 bit environment, a 1.0 should be 0b0_011111111111111_0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
with the next smaller one being 0b0_011111111111110_1111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111
(underscores after 1 and 15 bits to separate components).
Using nextafter however, I get
0b0_011111111111110_1111111111111111111111111111111111111111111111111111000000000000000000000000000000000000000000000000000000000000
which seems like it totally ignores the lower 60 bits.
The arithmetic itself does work just fine with 128 bits: subtracting the 1 and the next smaller number (with the bit pattern manually specified) does give me the same answer as the 128 bit cfloat.
This also happens when I want to set a variable to 1.3 which should be a repeating pattern, but again, the lowest 60 bits are blank.

As for the failing CI test, this is the result of reportingTestCases:
```
  10   │ posit truncate function validation: report test cases
  11   │ trunc trunc                       nar !=                     65536 reference value is                         0 0b0.00000.. vs 0b0.11111..
  12   │ trunc                                                        posit<6,2>() FAIL 1 failed test cases
  13   │ round round nar                       != 65536                     reference value is 0                         0b0.00000.. vs 0b0.11111..
  14   │ round                                                        posit<6,2>() FAIL 1 failed test cases
  15   │ floor floor nar                       != 65536                     reference value is 0                         0b0.00000.. vs 0b0.11111..
  16   │ floor                                                        posit<6,2>() FAIL 1 failed test cases
  17   │ ceil ceil nar                       != 65536                     reference value is 0                         0b0.00000.. vs 0b0.11111..
  18   │ ceil                                                         posit<6,2>() FAIL 1 failed test cases
  19   │ posit truncate function validation: FAIL
```

So, if I read this right, only one case (per function) fails and that is NaR. Could it be that the conversion of the NaR to a float (or the subsequent floor/ceil/etc) is handled differently on (some) ARM CPUs? You said it worked on the M1 though... I'll try to look into this.

OTOH: I would assume that converting a posit into a float is not related to this PR. It seems like to_float delegates to to_double which extracts the posit components and turns them into a double which is then casted to a float, so there is no long double involved.

yboettcher:
When I try to just convert a NaN into an int, I get different results on the two platforms. No posits or anything involved, just pure C++ floats and ints:

```cpp

int main() {
    // sanity check
    assert(sizeof(float) == sizeof(int));

     // nan bit pattern
    int nan_pattern = 0x7fc00000;

     // create float from pattern
    float nan_float;
    std::memcpy(&nan_float, &nan_pattern, sizeof(float));

    // use normal int conversion on the floor
    int result = int(nan_float);

    // check results
    std::cout << "float: " << nan_float << std::endl;

    std::cout << "result: " << result << std::endl;
    std::cout << "result_pattern: " << std::bitset<32>(result) << std::endl;

    return 0;
}
```
Gives me this output on x86:
> float: nan
result: 1970496882
result_pattern: 11011010010010100010111111010011


and this on my raspberry pi:
> float: nan
result: 8
result_pattern: 01100111001100000001000000000000

Am I right in assuming that the result of converting a nan into an integer type is just undefined and behaves differently on different platforms? What confuses me as well is that the bit pattern does not match the printed result...

I found this question https://stackoverflow.com/questions/10366485/problems-casting-nan-floats-to-int, which has some interesting information:
> if the floating value is infinite or NaN or if the integral part of the floating value exceeds the range of the integer type, then the "invalid" floating- point exception is raised and the resulting value is unspecified.

and
> Intel specifies instruction behavior very carefully, and the behavior for converting a floating-point NaN to a 32-bit integer is to return 0x80000000

The first statement would explain what I am seeing here, the second one however leads me to question my test code above. I am definitely not getting 0x80000000 on the x86. I did get that earlier though (when converting into long I think).

In any case, I think this is quite far from the actual topic of the PR and has nothing to do with quad precision floating point numbers.

Ravenwater:
@yboettcher I have found that the NaN and Inf patterns are different among compilers, so this trick to push bits into a float storage might conflict with how the C++ runtime you are using is interpreting the patterns.

Ravenwater:
> First up: I now got your example from [#336 (comment)](https://github.com/stillwater-sc/universal/pull/336#issuecomment-1527611965) to succeed. I just messed up the ordering of source and destination for memcpy. The CI test is still failing though, so I would assume that it is unrelated to the component extraction. This concerns me a little though. As wrong as my previous implementation was, I am surprised that none of the tests actually caught this.
> 
> I actually ran into some other strange behaviour when trying out the `nextafter` function. In a 128 bit environment, a 1.0 should be 0b0_011111111111111_0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000 with the next smaller one being 0b0_011111111111110_1111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111 (underscores after 1 and 15 bits to separate components). Using nextafter however, I get 0b0_011111111111110_1111111111111111111111111111111111111111111111111111000000000000000000000000000000000000000000000000000000000000 which seems like it totally ignores the lower 60 bits. The arithmetic itself does work just fine with 128 bits: subtracting the 1 and the next smaller number (with the bit pattern manually specified) does give me the same answer as the 128 bit cfloat. This also happens when I want to set a variable to 1.3 which should be a repeating pattern, but again, the lowest 60 bits are blank.
> 
> As for the failing CI test, this is the result of reportingTestCases:
> 
> ```
>   10   │ posit truncate function validation: report test cases
>   11   │ trunc trunc                       nar !=                     65536 reference value is                         0 0b0.00000.. vs 0b0.11111..
>   12   │ trunc                                                        posit<6,2>() FAIL 1 failed test cases
>   13   │ round round nar                       != 65536                     reference value is 0                         0b0.00000.. vs 0b0.11111..
>   14   │ round                                                        posit<6,2>() FAIL 1 failed test cases
>   15   │ floor floor nar                       != 65536                     reference value is 0                         0b0.00000.. vs 0b0.11111..
>   16   │ floor                                                        posit<6,2>() FAIL 1 failed test cases
>   17   │ ceil ceil nar                       != 65536                     reference value is 0                         0b0.00000.. vs 0b0.11111..
>   18   │ ceil                                                         posit<6,2>() FAIL 1 failed test cases
>   19   │ posit truncate function validation: FAIL
> ```
> 
> So, if I read this right, only one case (per function) fails and that is NaR. Could it be that the conversion of the NaR to a float (or the subsequent floor/ceil/etc) is handled differently on (some) ARM CPUs? You said it worked on the M1 though... I'll try to look into this.
> 
> OTOH: I would assume that converting a posit into a float is not related to this PR. It seems like to_float delegates to to_double which extracts the posit components and turns them into a double which is then casted to a float, so there is no long double involved.

The fact that the result is 65536 or 0 is interesting. That feels like there is an improper mask at play. The posit API has a setnar() method, so it would be interesting to see this code on your Arduino:

```
   posit<6,2> p;
   p.setnar();
   ReportValue(p);     // comes from #include <universal/verification/test_suite.hpp>
```

Ravenwater:
@yboettcher you found a bug in the posit printing of binary format. I put in the fix in this commit: https://github.com/stillwater-sc/universal/commit/40ef3ad2a4729d08a455ac98496f31608f61a4b3

Ravenwater (approved):
LGTM

PR #407, opened 2024-01-18
V3.75: specialized posit enhancements

Ravenwater:
Restructuring the posit regression tests and adding a fast posit<16,2>


theo-lemurian (approved):
passes full regression suite

Issue #343, opened 2023-06-02
Specialization of standard posits


RaulMurillo:
Since the latest standard for posits (https://github.com/posit-standard/Posit-Standard-Community-Feedback) fixes the exponent size to 2 bits, it would be desirable to have fast specialization of posit types with 2 exponent bits (at least, 8, 16, 32, and even 64 bits) for performance in software experimentation.
I don't know how difficult it would be to adapt the current specialized/posit_8_0.hpp and specialized/posit_16_1.hpp (specialized/posit_32_2.hpp will be the same I guess) to the new standard.

RaulMurillo:
> Would you be able to drive this development?
 
Sure! As far as I know, it should resemble previous specialized implementations, with adjustments to the parameters and arithmetic operations for those specific formats. Am I right?

Also, could you please point me to some regression tests to ensure that the specialized implementations provide the same functionality as the default ones?

theo-lemurian:
I can set up the regression tests for the standard specialized posits so
you can focus on the implementation

Ravenwater:
@RaulMurillo I have created a branch 'fast_specialized_posit', and enabled the fast posit path in the regressions. I put a skeleton together for posit<16,2> and that yielded this:

```
Fast specialization posit<16,2> configuration tests
 posit< 16,2> useed scale     4     minpos scale        -56     maxpos scale         56  :  0

Special case tests
posit< 16, 2> Initialize to zero:            PASS
posit< 16, 2> Initialize to NAN              PASS
posit< 16, 2> Initialize to INFINITY         PASS
posit< 16, 2> sign is true                   PASS
posit< 16, 2> is negative                    PASS
posit< 16, 2> sign is false                  PASS
posit< 16, 2> is positive                    PASS
posit< 16, 2>                                                addition       PASS
posit< 16, 2>                                                subtraction    PASS
posit< 16, 2>                                                multiplication PASS
posit< 16, 2>                                                division       PASS
```

For REGRESSION_LEVEL_1. Levels 2, 3, and 4 will add logic, randoms, and the math library to the mix. I also enabled it in the CI, so any check-ins against that branch will automatically run the regression suite.

I can show you how to work with the REGRESSION levels and the manual test regression infrastructure.

[...]

PR #409, opened 2024-02-02
Switched to use posito when testing specialized posit<16,2>

davidmallasen:
Following issue #343. Everything seems to be working fine, but I still get different results in my application. Maybe the problem is not in the basic operations but in some other detail.

### DIFF
--- a/static/posit/specialized/posit_16_2.cpp
+++ b/static/posit/specialized/posit_16_2.cpp
@@ -12,9 +12,10 @@
 // second: enable posit arithmetic exceptions
 #define POSIT_THROW_ARITHMETIC_EXCEPTION 1
 #include <universal/number/posit/posit.hpp>
+#include <universal/number/posito/posito.hpp>
 #include <universal/verification/posit_test_suite.hpp>
 #include <universal/verification/posit_test_suite_mathlib.hpp>
-#include <universal/verification/posit_test_suite_randoms.hpp>
+#include <universal/verification/posit_specialized_test_suite_randoms.hpp>
 
 // Standard posit with nbits = 16 have es = 2 exponent bit.
 
@@ -143,6 +144,8 @@ epilog:
 	test = "is positive";
 	nrOfFailedTestCases += ReportCheck(tag, test, p.ispos());
 
+	std::cout << "Basic arithmetic tests " << RND_TEST_CASES << " randoms each\n";
+	std::cout << "Using the posit oracle posito as reference\n";
 	nrOfFailedTestCases += ReportTestResult(VerifyBinaryOperatorThroughRandoms<nbits, es>(reportTestCases, OPCODE_ADD, RND_TEST_CASES), tag, "addition      ");
 	nrOfFailedTestCases += ReportTestResult(VerifyBinaryOperatorThroughRandoms<nbits, es>(reportTestCases, OPCODE_SUB, RND_TEST_CASES), tag, "subtraction   ");
 	nrOfFailedTestCases += ReportTestResult(VerifyBinaryOperatorThroughRandoms<nbits, es>(reportTestCases, OPCODE_MUL, RND_TEST_CASES), tag, "multiplication");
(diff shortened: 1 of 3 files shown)


static/posit/specialized/posit_16_2.cpp
```diff
@@ -12,9 +12,10 @@
 // second: enable posit arithmetic exceptions
 #define POSIT_THROW_ARITHMETIC_EXCEPTION 1
 #include <universal/number/posit/posit.hpp>
+#include <universal/number/posito/posito.hpp>
 #include <universal/verification/posit_test_suite.hpp>
 #include <universal/verification/posit_test_suite_mathlib.hpp>
-#include <universal/verification/posit_test_suite_randoms.hpp>
```
Ravenwater, line 17: Let's not modify the regression suite, but instead, let's add a new exploratory test so we can RCA this issue

Ravenwater (changes requested):
Let's not modify the regression test, but instead let's create a new exploratory test in the posito section of the tree. That way we have much more freedom to experiment and we are not polluting the posit regression test suite.

We also should not need the new posit_specialized_test_suite_randoms.hpp or the test_reporters.hpp modification: that should all be parameterized so we can throw both posit<> and posito<> types at them.

So, recap: 
- posit_specialized_test_suite_randoms.hpp and test_reporters.hpp mods should be removed
- back out the changes to posit_16_2.cpp to its original form
- add a new test posit_16_2.cpp to the posito/arithmetic directory that will contain the code for RCAing any differences between posit specialized and posito.

static/posit/specialized/posit_16_2.cpp
davidmallasen: @Ravenwater I have a few issues with what you ask:
- How do you propose to do the type generalization in `posit_test_suite_randoms.hpp`? As it is now it's not straightforward to do, and I'm not sure what your plans are.
- To keep the `test_reporters.hpp` as-is, we'd need some sort of conversion between posit and posito. Where should that be added? I guestt it should suffice by copying the bits, but I might be wrong.
- Why not keep these extra tests in the `posit/specialized` folder? That's where the tests should be, not in some posito folder since we're not testing for posito. I'm saying this especially because these tests are passing. It also serves as a way of avoiding the issue we had at the beginning where the specialized posit<16,2> was behaving as a posit<16,1> and still passing these tests.

ghost: @davidmallasen you are correct, we want to bring in posito<> as an oracle. 

Your mod got caught in the work-in-progress of unifying the regression API among all the number systems, which tripped over SFINAE failures. I am going to take your design and merge it into the new API and try to make it do double duty between marshalling through double and using an oracle.

davidmallasen: Perfect! Feel free to use my modifications in this PR in any way you are designing the new API. If at any point this PR is no longer relevant we can close it.

Ravenwater:
@davidmallasen I got the new api in branch v3.76

But as I was prepping for the merge, I had new thought: 
1- randoms are unlikely to catch corner cases
2- we tested the arithmetic operators exhaustively
3-the regression test shows float conversion failures

From those observations, it appears that our RCA between behavioral differences between fast posit<16,2> and posito<16,2> should first resolve the float conversion failures.

Ravenwater:
And the saga continues:

```
posit conversion validation: results only
posit<5,2>                                                   conversion PASS
posit<6,2>                                                   conversion PASS
posit<7,2>                                                   conversion PASS
posit<8,2>                                                   conversion FAIL 484 failed test cases
posito<8,2>                                                  conversion PASS
posit<9,2>                                                   conversion PASS
FAIL = 0.06251519918             did not convert to 0.06253051758             instead it yielded  0.0625                     raw 0b0.01.00.00000000000
FAIL = 0.999878943               did not convert to 0.9997558594              instead it yielded  1                          raw 0b0.10.00.00000000000
posit<16,2>                                                  conversion FAIL 2 failed test cases
FAIL = 0.06251519918             did not convert to 0.06253051758             instead it yielded  0.0625                     raw 0b0.01.00.00000000000
FAIL = 0.999878943               did not convert to 0.9997558594              instead it yielded  1                          raw 0b0.10.00.00000000000
posito<16,2>                                                 conversion FAIL 2 failed test cases
posit conversion validation: FAIL
```

Looks like there is a bad cast assumption going on in the stack.

davidmallasen:
Hello @Ravenwater. For a generic Oracle type, I don't think the Oracle types should depend on the type under test. For this I would use something like MPFR, although I don't know how complicated it would be to include in Universal. However, in the context of the specialized posit, we do need some way to avoid what happened between posit<16,1> and posit<16,2>, so comparing the bit strings with the stable posit bitblock implementation would still be the way to go.

Thanks for your time on this, and great that you found where the problem is! I didn't get those errors when launching the tests the previous time I don't know why. That could be the problem that I saw where the results were not exactly the same but differed a bit between the specialized/non-specialized versions. Would this be some error in https://github.com/stillwater-sc/universal/blob/60fe91ca69edcbbd4386542e80fbf62e623ad09e/include/universal/number/posit/specialized/posit_16_2.hpp#L696 ?

Ravenwater:
@davidmallasen :-) Nothing to do with either rounding or casting in the posit implementation: it was a bug in the regression test:

```
posit conversion validation: results only
posit<5,2>                                                   conversion PASS
posit<6,2>                                                   conversion PASS
posit<7,2>                                                   conversion PASS
posit<8,2>                                                   conversion FAIL 484 failed test cases
posito<8,2>                                                  conversion PASS
posit<9,2>                                                   conversion PASS
posit<16,2>                                                  conversion PASS
posito<16,2>                                                 conversion PASS
```

the specialized posit<8,2> does use a different conversion algorithm for the sake of speed, and that still needs some TLC.

But that leaves still a hole in our theory what to test to RCA the difference between posit<16,2> and specialized posit<16,2> in your application.

Any other ideas? Can you describe the computational path you are seeing differences in?

Ravenwater:
@davidmallasen and concerning an Oracle type, we are working on a fast Priest arbitrary precision type a la MPFR that could be utilized across Universal. One of the requirements of Universal is that it has no external dependencies, and thus linking in a static library like MPFR is a non-starter.

davidmallasen:
@Ravenwater I'll have to try to pinpoint where the two operations converge. When I get some time I'll try to narrow it down a bit so I can share where the discrepancy is.
I imagined adding MPFR would be a no-go 👍🏽

PR #415, opened 2024-03-23
PLU update

jamesquinlan:
Here is a PLU update. I artificially made my test matrices to include the permutation, and the bug in the permutation has been fixed.


theo-lemurian:
code changed had already trickled through, closing PR

PR #426, opened 2024-08-07
V3.78: adding a double-double arithmetic type

Ravenwater:
adding the double-double arithmetic type for fast higher precision than double floating-point calculations


jamesquinlan (approved):
LGTM.

PR #452, opened 2025-04-18
V3.81 - adding statistical comparison function for closure plots

C-Wirth:
## Files added
evaluate_closure_of_number_systens.hpp - a header file utility that generates the statistics associated with a closure plots
evaluate_small_number_systems.cpp - generates the statistics for closure plots for 4,6,8,10 and 12 bit cfloats, posits and lns with comparable dynamic ranges.


include/universal/utility/evaluate_closure_of_number_systems.hpp
```diff
@@ -0,0 +1,202 @@
...
+// ****************************************************************
+// 
+// This file is part of the universal numbers project, which is released under an MIT Open Source license.
+
+#include <iostream>
+#include <string>
+#include <string_view>
+#include <fstream>
+
+#include <universal/number/posit/posit.hpp>
```
Ravenwater: the posit/cfloat/lns includes should not be in this include file. 

the posit/cfloat/lns include files bring in those number systems and you are not using them 'specifically' in this file, so they should not be included. For include files, you should always strive to have a minimum set of includes that make the include file usable. In templated libs like universal, an include typically declares a set of parameterized classes and/or functions, that SHOULD be able to work with any of the types in the library. But you do not need to include them, as a matter of fact, you shouldn't include them. That is up to the calling application to bring together a context in which say a posit<> type is interpretable. But you need to leave them out of the defining include files like this.

include/universal/utility/evaluate_closure_of_number_systems.hpp
```diff
@@ -0,0 +1,202 @@
...
+
+// the data for each operation in a buildClosurePlots is stored in this struct
+struct operationResults{
+    float TotalOps;
+    std::string Exact;
+    std::string Approximate;
+    std::string Overflow;
+    std::string Underflow;
+    std::string Saturate;
+    std::string narOrNan;
```
Ravenwater: narOrNan has a different camelcase form than the other vars, which jumps out at me. Secondly, the name narOrNan is clearly an 'unclean' concept, even the name can't figure out what it stands for. This is an architectural design decision, so needs careful analysis as it needs to be unambiguous in its purpose.

Personally, Not A Real, is not as widespread in use as Not A Number. John's argument is that NaN is not kosher because the value is being used as a number. And his improvement to Not A Real was to resolve that tension. Problem is that it has not been picked up, and there is more familiarity with NaN.

You can make the call what you would like to call it, but make it ONE concept (so remove the or) and name it with the same CamelCase of the other vars.

include/universal/utility/evaluate_closure_of_number_systems.hpp
```diff
@@ -0,0 +1,202 @@
...
+///<summary>
+/// helper function for processASystem()
+/// formats and prints the results
+/// /</summary>
+///
+///<param> name="results" a struct that stores all of the statistics of the closure plot for a number system </param>
+/// <returns> void </returns>
+/// 
+void printResults(const storeResults& results) {
+    constexpr unsigned data_width = 15;
```
Ravenwater: Excellent, that is a good use of constexpr. 

I have gotten into the habit to using CAPITAL LETTERS for constants, so that you 'know' that we are dealing with a constant. But I am not consistent with it, so you will see constexpr names like bTrace.

for COL_WIDTH, and DATA_WIDTH, it makes sense to me. I leave it to your decision if you rename it or not.

playground/evaluate_small_number_systems.cpp
```diff
@@ -0,0 +1,164 @@
...
+            cout << "\n";
+        }
+
+        if(_12bit){
+            cout << "Processing 12 bit systems:\n\n" ;
+            processASystem<cfloatConfigs::cfg12_t>(cfloatConfigs::cfg12_str);
+            processASystem<positConfigs::cfg12_t>(positConfigs::cfg12_str);
+            processASystem<lnsConfigs::cfg12_t>(lnsConfigs::cfg12_str);
+            cout << "\n";
+        }
```
Ravenwater, line 162: This is another architecture question. One of the great powers of template meta programming is that you can 'specialize' the class or function to a particular context, such as 4bit, 6bit, 8bit, etc. So, when I see code like if(_4bit){} I immediately ask the question, could you have done that with meta programming? What is the specialization that makes processASystem<> adapt to 4bits versus 10bits? 

One trick is to use a so-called tag, which is typically a struct that allows you to communicate dynamic information. Metaprogramming needs something to key off the specialization, so names structs or enums are good candidates for this trick.

I leave the decision to you if you want to try to design a metaprogramming design that removes the if(_##bits) structure and replaces that with a single invokation. It is an opportunity to try a new programming trick, but it is likely to take you a couple of weeks to get it done, so I won't blame you if you say no.

Ravenwater (changes requested):
couple of comments to improve the code.

playground/evaluate_small_number_systems.cpp
C-Wirth: Let's discuss this at the next meeting.  For now, I honestly don't have the bandwith for a couple-of-week learning period to optimize the code.  We can leave it for now, and revisit it another day.

Ravenwater (approved):
LGTM

PR #454, opened 2025-07-17
added rpoly root-finding

fivekoi:
* Added the Jenkin's Traub real polynomial root finding algorithm under mixedprecision/roots.

### DIFF
--- a/mixedprecision/roots/rpoly.cpp
+++ b/mixedprecision/roots/rpoly.cpp
@@ -76,7 +76,7 @@ inline std::complex<Real> EvaluateComplexPolynomial(const sw::universal::blas::v
 template <typename Real>
 sw::universal::blas::vector<Real> RemoveLeadingZeros(const sw::universal::blas::vector<Real>& polynomial_in) {
   size_t i = 0;
-  while (i < (polynomial_in.size() - 1) && polynomial_in[i] == 0.0) {
+  while (i < (polynomial_in.size() - 1) && polynomial_in[i] == Real(0.0)) {
     ++i;
   }
   return Tail(polynomial_in, polynomial_in.size() - i);
@@ -448,7 +448,7 @@ bool JenkinsTraubSolver<Real>::ExtractRoots() {
     // Remove any leading zeros of the polynomial.
     polynomial_ = RemoveLeadingZeros(polynomial_);
     // Normalize the polynomial.
-    polynomial_ /= polynomial_[0];
+    polynomial_ = polynomial_ / polynomial_[0]; // NOTE: polynomial /= polynomial_[0] does not work...
     const int degree = polynomial_.size() - 1;
 
     // Allocate the output roots.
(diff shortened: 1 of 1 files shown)


mixedprecision/roots/rpoly.cpp
```diff
@@ -0,0 +1,1052 @@
...
+    size_t startIdx = vector.size() - size > 0 ? vector.size() - size : 0;
+    for (size_t i = startIdx; i < vector.size(); ++i){
+        tail[i] = vector[i - startIdx];
+    }
+
+    return tail;
+}
+
+template <typename Real>
+void SetZero(sw::universal::blas::vector<Real>& vector, size_t size){
```
Ravenwater: this semantic is in the library

using Real = posit<32,2>;
using Vector = blas::vector<Real>;

Vector v(7341, 0);  // construct size and set elements to value

Vector w(16);

w = Real(0);  // this will broadcast the value across the vector elements

mixedprecision/roots/rpoly.cpp
```diff
@@ -0,0 +1,1052 @@
...
+    LINEAR_CONVERGENCE = 1,
+    QUADRATIC_CONVERGENCE = 2
+};
+
+// Evaluate the polynomial at x using the Horner scheme.
+template <typename Real>
+inline Real EvaluatePolynomial(const sw::universal::blas::vector<Real>& polynomial, const Real& x) {
+    Real v = 0.0;
+    for (size_t i = 0; i < polynomial.size(); ++i) {
+        v = v * x + polynomial[i]; // TODO: quire?
```
Ravenwater: v * x + p[i]  has two rounding events: the multiply and the add. You can reduce that to one with the quire, but when you do the assignment, you would lose the small values again.

the quire only works for sum of products, but here we have really have a sequence of products and the quire can't resolve that.

Order of the computation is the only dial you have to reduce catastrophic cancellation.

mixedprecision/roots/rpoly.cpp
```diff
@@ -0,0 +1,1052 @@
...
+
+    // If the quotient is a constant then polynomial is degree 2 and the math is simple.
+    if (quotient.size() == 1) {
+        // TODO: maybe this could be written cleaner??
+        remainder = {polynomial[1] - polynomial[0] * quadratic_divisor[1], polynomial[2] - polynomial[0] * quadratic_divisor[2]};
+        return; 
+    }
+
+    quotient[1] = polynomial[1] - polynomial[0] * quadratic_divisor[1];
+    for (size_t i = 2; i < polynomial.size() - 2; i++) {
```
Ravenwater: remember, ++i does not need a temporary, i++ does, so ++i is the preferred from in the loop spec

Ravenwater (changes requested):
just a couple of minor edits.

Ravenwater (approved):

PR #474, opened 2025-08-20
Mark blfloat non-tmpl functions as inline

omegacoleman:
This PR fixes ODR linking issues with multiple TU:

```
..universal/number/bfloat/math/pow.hpp:11: multiple definition of `sw::universal::pow(sw::universal::bfloat16, sw::universal::bfloat16)';..
..universal/number/bfloat/math/pow.hpp:16: multiple definition of `sw::universal::pow(sw::universal::bfloat16, int)';..
```

.. and many more.

The last PR baffled me as the v3.83 branch is ahead of main. So I assume this is where the development is going on. If you want a PR to main, I can do that too.

### DIFF
--- a/include/sw/universal/number/bfloat/bfloat16_fwd.hpp
+++ b/include/sw/universal/number/bfloat/bfloat16_fwd.hpp
@@ -10,6 +10,6 @@ namespace sw { namespace universal {
 
 // forward references
 class bfloat16;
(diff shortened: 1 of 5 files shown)


Ravenwater:
That pattern we'll need to repeat on the rest of the library. I'll take care of that.

PR #480, opened 2025-08-28
Added CI script for running Test Suite on RISC-V

singul4ri7y:
- Added `riscv-tests.yml` GitHub workflow script in `.github/workflows/` enabling RISC-V testing on a static QEMU RISC-V64 VM via cross-compilation.
 - Parallelized the tests for overall faster execution.
 - Also parallelized test execution for all the other platforms.


singul4ri7y:
Implemented in v3.97.

PR #492, opened 2026-01-19
UBSan / ASan fixes uncovered by intermittent test failures

loren-osborn:
## Summary

This PR addresses a collection of **UndefinedBehaviorSanitizer (UBSan)** and
**AddressSanitizer (ASan)** issues uncovered while investigating an
**intermittent test failure observed under Clang on x86** (approximately
10–20% failure rate).

While Clang/x86 made the problem easier to reproduce, **none of the issues
fixed here are Clang- or x86-specific**. They represent genuine instances of
undefined behavior that may manifest differently depending on compiler,
architecture, or optimization level.

---

## Scope

- Fixes multiple instances of **undefined behavior** and **out-of-bounds access**
- Improves runtime stability and determinism across toolchains
- Removes sanitizer-reported UB even when it did not consistently manifest as a test failure
- Does **not** change public APIs or numerical intent

---

### Grisu Algorithm (Not Fixed Here)

While working through UBSan reports, it became clear (with assistance from ChatGPT)
that the **Grisu algorithm as currently implemented is incorrect**.

- This PR **does not attempt to fix or redesign the Grisu algorithm**
- The issue is **explicitly documented in comments**
- The fixes here are limited strictly to removing UB and memory-safety violations

Correcting the Grisu implementation itself should be handled as a **separate,
algorithm-focused change**.

---

## Commit Structure (For Easier Review)

This PR is intentionally split into **three commits**:

1. **General UBSan / ASan fixes (manual)**
   - Straightforward UB fixes identified and corrected by hand

2. **Grisu-related UBSan fixes (manual)**
   - Removes UB in Grisu-related code paths
   - Adds documentation noting that the current algorithm is incorrect
   - No attempt is made to “fix” the algorithm itself

3. **UBSan / ASan fixes generated by OpenAI Codex**
   - Remaining sanitizer issues resolved by Codex
   - Changes appear correct and tests pass, but these commits
     **would benefit from additional human review**
   - Kept separate intentionally for auditability

---

## Testing

- Full test suite run with UBSan and ASan enabled
- The intermittent failures that triggered this investigation
  are no longer observed once these fixes are applied
- Remaining sanitizer-related issues (if any) are believed to be
  independent and are tracked separately

---

## Related Work

- A separate PR addresses **compiler-specific Clang x86 behavior**
- That PR focuses on consistent test failures, while this one focuses on
  **toolchain-agnostic undefined behavior**

### DIFF
--- a/include/sw/universal/number/fixpnt/fixpnt_impl.hpp
+++ b/include/sw/universal/number/fixpnt/fixpnt_impl.hpp
@@ -22,12 +22,12 @@
 
 /*
 The fixed-point arithmetic can be configured to:
-- throw exception on overflow
+- throw an exception on overflow
 - saturation arithmetic: saturate on overflow
 - modular arithmetic: quietly overflow into modular values
 
 The quietly overflow configuration is reasonable when you are using
-a fixed-point size that captures the dynamic of your computation.
+a fixed-point size that captures the dynamics of your computation.
 Due to the fact that no special cases are required, the arithmetic
 operators will be much faster than saturation.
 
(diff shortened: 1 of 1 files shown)


include/sw/universal/number/fixpnt/fixpnt_impl.hpp
Ravenwater, line 1: if (v == -v) is to check for the 0x100000 signed int case, which is a special case.

if (v == std::numeric_limits<Arith>::min()) is not correct as Arith is Modulo or Saturating, so it doesn't have a min().

Ravenwater (changes requested):
the fixpnt test I don't understand why it made that suggestion.

That is a case I have been bitten by a thousand times, so I thought it was a neat trick to identify that special case. 

Do you think it is equivalent? I do not think it is

PR #591, opened 2026-03-17
fix(cmake): Fix GCC build on Intel macOS (build failure, warnings, and small-angle std::cos(float) issue)

loren-osborn:
This PR fixes GCC support on Intel macOS, bringing the project back to a clean **build + test pass** state.

It consists of three focused commits:

---

### **1. Fix GCC build failure (Intel macOS)**

Resolves a blocking issue that prevented the project from compiling under GCC on Intel macOS.

---

### **2. Fix GCC warnings (Intel macOS)**

Addresses several compiler warnings encountered during the GCC build.

While resolving these, a couple of pre-existing areas were identified for future review (not modified in this PR):

* `include/sw/universal/number/unum2/unum2_impl.hpp:412`
  *(mixed-domain `shift_count`)*
* `include/sw/universal/number/unum2/lattice.hpp:122`
  *(suspicious `_N >> 1` vs `_N_half` logic)*

These are noted as follow-up items rather than being changed here to keep the diff minimal.

---

### **3. Generalize workaround for small-angle `std::cos(float)` issue**

Previously, there was a Clang/x86-specific workaround for incorrect behavior in `std::cos(float)` at very small input values.

This PR replaces that with a **generalized, evidence-based fix**:

* Detects at runtime (once) whether the float cosine path is unreliable on the current toolchain
* Falls back to the double implementation **only when needed**
* Avoids a blanket “always use double” approach
* Covers both:

  * GCC on Intel macOS
  * existing Clang/x86 behavior

This keeps the fix **narrow, portable, and performance-conscious**.

---

## **Validation**

Tested with:

```bash
make COMPILER=gcc test
```

Result:

* ✅ Builds successfully on GCC (Intel macOS)
* ✅ All tests pass

---

## **Notes**

* Changes are intentionally scoped to the minimum needed to restore correctness and stability.
* Further cleanup of flagged technical debt can be handled in follow-up PRs.
* The cosine fix is designed to be robust across toolchains without introducing unnecessary overhead.

## Summary by CodeRabbit

* **New Features**
  * Improved runtime selection for bfloat16 trigonometric computations to increase reliability.

* **Bug Fixes**
  * Added safer cleanup on memory allocation failures.

* **Refactor**
  * Made numeric types and indexing more consistent for 64-bit paths.
  * Ensured local numeric objects are value-initialized (safer defaults).
  * Adjusted how certain floating-point significand values are computed and displayed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -886,6 +886,9 @@ if(UNIVERSAL_BUILD_CI_LITE)
 	# conversions
 	set(UNIVERSAL_BUILD_NUMBER_CONVERSIONS ON)
 
+	# linear algebra data serialization
+	set(UNIVERSAL_BUILD_LINEAR_ALGEBRA_DATA ON)
+
 	# build the C API library
 	set(UNIVERSAL_BUILD_C_API_PURE_LIB ON)
 	set(UNIVERSAL_BUILD_C_API_SHIM_LIB ON)
--- a/include/sw/universal/internal/blockbinary/blockbinary.hpp
+++ b/include/sw/universal/internal/blockbinary/blockbinary.hpp
@@ -607,15 +607,30 @@ public:
 	constexpr void set(unsigned i) noexcept {	setbit(i, true); }
 	constexpr void reset(unsigned i) noexcept { setbit(i, false); }
 	constexpr void setbit(unsigned i, bool v = true) noexcept {
+		if (i >= nbits) return;
 		unsigned blockIndex = i / bitsInBlock;
 		if (blockIndex < nrBlocks) {
+			// GCC -O3 produces false-positive -Warray-bounds when inlining
+			// setbit across blockbinary instantiations with different nbits
+			// (e.g., blockbinary<129> calling into blockbinary<128> context).
+			// The bounds checks above are correct but GCC's interprocedural
+			// analysis loses track of which instantiation's _block is accessed.
+#if defined(__GNUC__) && !defined(__clang__)
+#pragma GCC diagnostic push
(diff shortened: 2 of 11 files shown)


loren-osborn:
"fix(cmake):" was the closest prefix I could find to "fix(build): "

include/sw/universal/number/bfloat16/math/functions/trigonometry.hpp
```diff
@@ -9,6 +9,35 @@
...
+	return bfloat16(std::cos(float(x)));
+}
+
+inline bfloat16 bfloat16_cos_double_path(bfloat16 x) {
+	return bfloat16(std::cos(double(x)));
+}
+
+inline bool bfloat16_cos_float_path_is_unreliable() {
+	static const bool needs_workaround = [] {
+		constexpr float probe = 0x1p-12f;
```
loren-osborn: I should have probably commented these better:
* `probe` - is a known problematic input value taken from Universal "bfloat16_mathlib" test
* `c` - the cosine(`probe`) computed by the problematic function
* `cd` - `c` cast to a double
* `s_est` - estimated (quadrant 1) sine(`probe`), calculated from `cd`
* `threshold_scale` - by what factor `s_est` is supposed to be correct.

Why do we want the computed sine(`probe`)?
Because for very small $x$, $\sin{x} \approx x$

I think line 35 is wrong... it should be closer to:

```
return ((std::abs(s_est) / threshold_scale) > std::abs(probe)) || ((std::abs(s_est) * threshold_scale) < std::abs(probe));
```

So, the corrected test above should tell us: is our estimated sine more than 8 times too big, or less than 8 times too small.

PR #614, opened 2026-03-23
fix(cfloat): supernormal and rounding bugs

ibriq:
## Summary

  Fixes three bugs in cfloat that affect supernormal configurations (hasMaxExpValues=true) and integer-to-cfloat rounding for small formats like FP8 E4M3. These were originally identified in 2024 while using the library to model FP8 hardware for a deep learning accelerator, and worked around locally at the time. This PR contributes proper fixes upstream.

 ### Bug 1: isinf() returns true for supernormal cfloats

When `hasMaxExpValues=true`, all max-exponent bit patterns are reclaimed as numeric values (supernormals) -- there is no infinity encoding. But `isinf()` was still pattern-matching those bit patterns as infinity. This caused downstream logic (comparisons, classification, formatted output) to misidentify valid supernormal values as infinite.

 **Fix:** `isinf()` now returns `false` immediately when `hasMaxExpValues=true`.

 ### Bug 2: Division by zero produces invalid infinity encoding for supernormal cfloats

 `operator/=` called `setinf()` for non-zero divided by zero, regardless of whether the type has an infinity encoding. For supernormal cfloats this wrote a bit pattern that represents a valid supernormal number, not infinity.

**Fix:** When `hasMaxExpValues=true`, non-zero/zero now produces `maxpos`/`maxneg` for saturating types, or quiet NaN for non-saturating types.

### Bug 3: Integer-to-cfloat rounding errors for small formats

Three related issues in `convert_unsigned/signed_integer` and its `round<>` helper:

- **Sticky bit mask off-by-one:** The mask used `(shift - 2)` instead of `(shift - 1)`, missing one bit position from the sticky calculation. For E4M3 with 32-bit int input, this produced incorrect round-to-even decisions (e.g., `101` rounded to `96` instead of `104`).
- **Rounding overflow left stale fraction bits:** When rounding overflowed the fraction (e.g., `0b111 + 1 = 0b1000`), `raw >>= 1` left non-zero bits in the fraction instead of  clearing to zero. This caused values at power-of-two boundaries to round incorrectly (e.g., `31` became `48` instead of `32`).
- **No overflow guard for out-of-range integers:** When the biased exponent exceeded the encoding range, the resulting bits overflowed into adjacent fields producing garbage.
Added a check that saturates to `maxpos`/`maxneg` or sets infinity as appropriate.

## Release Notes

* **Bug Fixes**
  * Refined floating-point rounding behavior for improved computational accuracy, particularly in overflow scenarios.

* **Tests**
  * Added regression testing for integer-to-floating-point conversion to ensure rounding accuracy across multiple configurations.

### DIFF
--- a/include/sw/universal/number/cfloat/cfloat_impl.hpp
+++ b/include/sw/universal/number/cfloat/cfloat_impl.hpp
@@ -2989,7 +2989,7 @@ protected:
 			bool round = (mask & raw);
 			if constexpr (shift > 1u) { // protect against a negative shift
 				StorageType allones(StorageType(~0));
-				mask = StorageType(allones << (shift - 2));
+				mask = StorageType(allones << (shift - 1));
 				mask = ~mask;
 			}
 			else {
@@ -3011,7 +3011,7 @@ protected:
 				if (round || sticky) ++raw;
 				if (raw == (1ull << fbits)) { // overflow
 					++exponent;
-					raw >>= 1u;
+					raw = 0;
 				}
 			}
 		}
--- a/static/float/cfloat/api/supernormal_and_rounding.cpp
+++ b/static/float/cfloat/api/supernormal_and_rounding.cpp
@@ -1,0 +1,206 @@
+// supernormal_and_rounding.cpp: regression tests for integer-to-cfloat
+// rounding accuracy
+//
+// Copyright (C) 2026 Stillwater Supercomputing, Inc.
+// SPDX-License-Identifier: MIT
+//
+// This file is part of the universal numbers project, which is released under
+// an MIT Open Source license.
+//
+// This test validates bug fixes in convert_unsigned/signed_integer's round<>:
+//
+// 1. The sticky bit mask used (shift-2) instead of (shift-1), missing one bit
+//    position from the sticky calculation. For E4M3 with 32-bit int input,
+//    this produced incorrect round-to-even decisions (e.g., integer 101
+//    rounded to 96 instead of 104).
+//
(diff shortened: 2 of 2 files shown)


include/sw/universal/number/cfloat/cfloat_impl.hpp
```diff
@@ -682,6 +682,19 @@ class cfloat {
 				// zero divide by zero yields quiet NaN (in MSVC it is labeled -nan(ind) for indeterminate)
 				setnan(NAN_TYPE_QUIET);
 			}
+			else if constexpr (hasMaxExpValues) {
+				// when max-exponent encodings are reclaimed as numeric values
+				// (supernormals), there is no infinity encoding, so we
```
Ravenwater: The assertion in the comment is incorrect. we have an NaN and an Inf encoding in the last two bit patterns in the range.

static/float/cfloat/api/special_cases.cpp
```diff
@@ -113,15 +113,20 @@ inline int TestInf() {
...
-	//std::cout << to_binary(r) << std::endl;
-	if (!r.isinf()) ++fails;
-	if (!r.isinf(sw::universal::INF_TYPE_NEGATIVE)) ++fails;
-	r = -r;
-	//std::cout << to_binary(r) << std::endl;
-	if (!r.isinf()) ++fails;
-	if (!r.isinf(sw::universal::INF_TYPE_POSITIVE)) ++fails;
-	r.setnan();
-	if (r.isinf()) ++fails;
+	if constexpr (hasMaxExpValues) {
```
Ravenwater: There are NaN and Inf encodings, so this change needs to be removed.

Issue #877, opened 2026-05-20
elreal Phase D: Comparison and sign determination


Ravenwater:
Part of #873. Blocked by Phase C.

## Scope

The signature operation of any lazy-real implementation: deciding the sign of a value (and from it, all binary comparisons) may require pulling arbitrarily many bits from the operand streams. This phase ships the comparison protocol *and* the bounded-budget policy that makes it usable.

## Deliverables

- `sign(const elreal&)` returning `-1 / 0 / +1` -- with a documented refinement budget and behavior at the budget ceiling
- `compare`, `operator<`, `operator<=`, `operator>`, `operator>=`
- `operator==` -- the algorithmically *undecidable* case in general; must take a tolerance / budget argument or return a `std::optional<bool>` if the budget is exhausted without resolving
- `is_zero()`, `is_positive()`, `is_negative()` -- convenience wrappers around `sign`
- Termination policy: either a global default budget or a per-call budget; document the choice. McCleeary discusses both options.
- `elastic/elreal/logic/*.cpp`

## Hints

- The "is `x` zero?" test is the canonical undecidable problem. The implementation must guarantee progress: each refinement step either narrows the interval enough to decide, or admits failure.
- This is the phase most directly responsible for honoring McCleeary's correctness proofs. Reference the dissertation theorems by section number in the implementation comments.

## Acceptance

- All comparisons terminate in finite time for any pair of `elreal` operands, either with a definite answer or with a documented "budget exhausted" outcome
- The budget ceiling is configurable
- Edge cases tested: comparing a value to itself, comparing two structurally-different streams that represent the same real, comparing to zero

Issue #876, opened 2026-05-20
elreal Phase C: Arithmetic -- +, -, *, /, unary


Ravenwater:
Part of #873. Blocked by Phase A.

## Scope

The four ring operations plus negation and `abs`. Each operator produces a new `elreal` whose k-th approximation is computed lazily from the operands' approximations.

## Deliverables

- `operator+`, `operator-`, `operator*`, `operator/` (and the `+= -= *= /=` siblings)
- `operator-()` unary negation
- `abs(const elreal&)` free function and member
- `elastic/elreal/arithmetic/{addition,subtraction,multiplication,division}.cpp`

## Algorithmic notes

- **Addition / subtraction**: directly compose the operand streams; interval bounds combine naturally. Catastrophic cancellation is one place where lazy refinement *helps* -- if the operand approximations don't bracket the sign of the result, the stream pulls more bits.
- **Multiplication**: the k-th approximation of `a * b` is *not* the product of the k-th approximations. Compute the operand precisions required to deliver k bits of result precision. McCleeary chapter 4 has the bookkeeping.
- **Division**: implement via Newton-Raphson on the reciprocal stream; quadratic convergence means each refinement step roughly doubles the bit count.
- **Termination guarantees**: division by a value that may be exactly zero is the canonical undecidable case. The implementation must either decide-or-bail with a documented budget, or throw an exception (guarded by `ELREAL_THROW_ARITHMETIC_EXCEPTION`).

## Acceptance

- Algebraic identities pass at full refinement: `(a + b) - b ~= a`, `(a * b) / b ~= a`, `a / a ~= 1`, etc., with tolerances scaled by the refinement budget (not by a fixed type epsilon)
- Cross-validation against `ereal` for inputs both can handle: at matched refinement targets, results agree within ULPs of that target

Issue #875, opened 2026-05-20
elreal Phase B: Construction from native types and strings


Ravenwater:
Part of #873. Blocked by Phase A.

## Scope

Construct `elreal` values from the standard input sources. No arithmetic yet -- only conversion-in.

## Deliverables

- Constructors from `double`, `float`, `long double` -- the lazy stream is trivial here: constant approximation at machine precision with zero refinement tail.
- Constructors from `signed char`, `short`, `int`, `long`, `long long`, and their unsigned counterparts -- integers are exact in IEEE-754 up to `2^53`, then the stream needs to encode the residual exactly.
- Constructor from a rational `p/q` (two integer inputs) -- the *interesting* exact case: the k-th approximation is `p/q` rounded to k bits. This is the canonical lazy-real construction example.
- Constructor from `std::string` for decimal/scientific literals (e.g. `"3.14159"`, `"6.022e23"`, `"1/3"`).
- Optional: cross-construction from `dd`, `qd`, `ereal<N>` (can defer to a follow-up if the lazy-stream representation makes this awkward).
- `elastic/elreal/conversion/*.cpp` -- one test file per source-type family.

## Hints

- For decimal literal parsing, `include/sw/universal/number/dfloat/dfloat_impl.hpp::parse()` handles the same families.
- For the rational constructor, the natural representation is: the k-th approximation rounds `p/q` to k bits via long division on the residual. McCleeary discusses this case explicitly.

## Acceptance

- All standard native types round-trip to within their representable precision
- `elreal("1/3")` is observably distinct from `elreal(1.0/3.0)` -- the rational constructor delivers more bits when asked for them
- Tests cover edge cases: signed zero, signed infinity (if `elreal` admits them; otherwise document the choice), NaN, subnormal doubles

Issue #874, opened 2026-05-20
elreal Phase A: Foundation -- lazy-stream representation and refinement protocol


Ravenwater:
Part of #873.

## Scope

Establish the type skeleton, storage representation, and refinement protocol for `elreal`. This phase ships nothing the user can call -- it establishes the substrate that B-I build on.

## Deliverables

- `include/sw/universal/number/elreal/elreal_fwd.hpp` -- forward declarations and any aliases
- `include/sw/universal/number/elreal/elreal_impl.hpp` -- main class skeleton with the storage layout (no arithmetic yet)
- `include/sw/universal/number/elreal/elreal.hpp` -- umbrella header
- `include/sw/universal/number/elreal/exceptions.hpp` -- exception hierarchy, guarded by `ELREAL_THROW_ARITHMETIC_EXCEPTION` (follow the pattern in `include/sw/universal/number/dfixpnt/exceptions.hpp`)
- `include/sw/universal/number/elreal/numeric_limits.hpp` -- `std::numeric_limits<elreal>` specialization (constants can stub for now; flesh out in Phase E)
- `include/sw/universal/number/elreal/manipulators.hpp` -- `type_tag`, `to_binary` stubs
- `include/sw/universal/traits/elreal_traits.hpp` -- `is_elreal` etc., parallel to existing traits files
- `elastic/elreal/api/api.cpp` -- one canary test that constructs an `elreal` and links cleanly

## Design questions to settle in this phase

- **Storage**: continuation-style closures (`std::function<approximation(int)>`), an explicit DAG of operation nodes, or a `std::vector` of memoized approximations with a "refine to k bits" entry point? McCleeary's dissertation favours the latter; verify against chapters 3-4 before committing.
- **Approximation element**: interval `(lo, hi)` double pair, center+radius double pair, or a multi-precision component (`floatcascade<2>` for higher-quality bounds)? Pick one and document the rationale in a header comment.
- **Refinement protocol signature**: `refine_to(precision_bits)` returning the next approximation, or `at(k)` returning the k-th approximation? Equivalent in expressiveness; the API shape affects every downstream phase.

## Acceptance

- `cmake -DUNIVERSAL_BUILD_NUMBER_ELREALS=ON ..` configures cleanly
- `api/api.cpp` builds and runs under both gcc and clang
- Storage representation, approximation element, and refinement protocol are documented in a header comment on `elreal_impl.hpp`

## Blocks

Phases B through I depend on this phase.

[...]

PR #886, opened 2026-05-21
feat(elreal): Phase D -- sign, compare, ordering with refinement budget

Ravenwater:
## Summary

Phase D of epic #873. Implements [#877](https://github.com/stillwater-sc/universal/issues/877).

Adds the comparison machinery: `sign()`, `compare()`, and the six ordering operators. The refinement budget is the *whole point* of Phase D — it makes the "undecidable in general" sign-determination problem tractable by giving the caller explicit control over how many components to materialise before declaring a value indistinguishable from zero.

## What changed

- **`inline constexpr std::size_t elreal_default_budget = 8;`** — ~424 bits cumulative, enough for any practical signed comparison.
- **`sign(v, budget)`** — walks the stream, returns sign of first non-zero component, or 0 if budget exhausted. Special-cases NaN (returns 0, unordered) and inf (returns ±1 without budget walk).
- **`compare(a, b, budget)`** — `sign(a - b, budget)`. A Phase I performance pass can fuse the operand walks.
- **`<`, `<=`, `>`, `>=`, `==`, `!=`** — all six operators. IEEE-754 NaN semantics: ordering operators all return false on NaN; `!=` returns true (yes, even `NaN != NaN`).

## Algorithm (McCleeary 2019, dissertation Sec. 4.2)

For a normalized elreal with components `c_0, c_1, ..., c_{N-1}` satisfying the non-overlapping property `|c_{i+1}| <= ulp(c_i)/2`, the sign of the true value equals the sign of the first non-zero component. The remaining components are corrections smaller than the leading non-zero by at least one bit each, so their sum cannot flip the sign.

So sign determination boils down to: walk components, return on first non-zero, give up at budget. Each operator path's generator preserves the non-overlapping property (the EFT primitives in Phase B/C are designed for this), so the algorithm is sound.

## The point of lazy arithmetic, on display

The acceptance criterion of #877 says:
> Comparing two structurally-different streams that represent the same real

`elreal(1, 3)` represents 1/3 *exactly* (Phase B promise). `elreal(1.0/3.0)` represents the IEEE-754 rounding of 1/3, which is *less than* 1/3 by the rational residual. At default budget, the comparison `elreal(1,3) > elreal(1.0/3.0)` returns true — the lazy refinement finds the rational residual at depth 1 of their difference. **This is the whole reason lazy real arithmetic exists.** Test in `compare.cpp` exercises this directly.

## Budget semantics

- `budget = 0` → no materialisation, sign returns 0
- `budget = 1` → only depth-0 component checked
- `budget = 2` → depth-0 and depth-1 components checked
- Default `8` → enough for any precision-sensitive comparison short of hash-collision-grade equality
- Callers needing more pass a larger budget explicitly

The "value is mathematically zero" case (e.g., `a - a`) always returns 0 regardless of budget — the all-zero stream is decided by component 0 plus the non-overlapping property propagation.

## Test plan

Three new binaries under `elastic/elreal/logic/`:

| Binary | Coverage |
|---|---|
| `elreal_sign` | positive/negative/zero, NaN unordered, refinement past depth 0 finds the rational residual, budget exhaustion on `a - a`, signed-zero |
| `elreal_compare` | `compare()` and all six operators, `a == a` (same object), `a == b` (equal doubles), **rational vs rounded inequality** (the lazy-real distinctness test), NaN semantics, inf comparisons |
| `elreal_budget` | `elreal_default_budget` constant contract, `budget = 0` returns 0, **`budget = 1` vs `budget = 2` on `(rat - rnd)`** (verifies that budget controls refinement depth) |

- [x] gcc 13: all 3 new + 9 prior tests PASS
- [x] clang: all 3 new + 9 prior tests PASS

## What does *not* land in this PR

- Refined `is_zero()` / `is_positive()` / `is_negative()` free functions — deferred. The existing `iszero()` member is the cheap pre-refinement check; users who want refined zero-detection use `sign(v) == 0`.
- A throw mode for "undecidable comparison" via `elreal_undecidable_comparison` — deferred. The exception type is defined (Phase A) but not currently raised; the budgeted-equality semantics are clean without it.
- Performance: `compare(a, b)` allocates an `elreal` for `a - b`. A fused walk that pulls bits from `a` and `b` in lockstep is Phase I.

## Related

- Epic: #873
- Phases A–C merged: #874, #875, #876 via PRs #883, #884, #885
- Phase D issue: #877

## Summary by CodeRabbit

* **New Features**
  * Added standard comparison operators and equality for elreal numbers
  * Implemented sign determination and a budgeted comparison mode for refinement-controlled ordering/equality

* **Tests**
  * Added test suites validating sign, comparison, NaN/inf semantics, and budgeted behavior

* **Documentation**
  * Updated phase documentation to reflect the new refinement/budgeted semantics

[Review Change Stack](https://app.coderabbit.ai/change-stack/stillwater-sc/universal/pull/886?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Ravenwater:
**Review status after CodeRabbit's first pass** (commit 0906c2e):

| Finding | Verdict | Action |
|---|---|---|
| `budget.cpp:32-36` only checks `elreal_default_budget < 2` (lower bound); a regression to a different default would pass silently | **valid** | Kept the `>= 2` semantic check (documents the minimum viable value) **and** added an explicit `== 8` contract check (locks the documented default). If the default is intentionally retuned later, this single line gets updated in the same PR -- catching silent regressions. |

Both assertions PASS on gcc 13 and clang.

Issue #906, opened 2026-05-21
elreal Phase L: Depth-2+ arithmetic refinement -- Newton for / and sqrt


Ravenwater:
Part of #903. Lifts the depth-1 ceiling on division and square root.

## Scope

As of Phase G, \`elreal::operator/\` materialises \`c0 = a.at(0) / b.at(0)\` (a single double divide) and returns a degenerate generator that produces zero for k >= 1. \`elreal::sqrt\` does the analogous shortcut for the leading component. This phase replaces both with Newton iteration that drives the result to any caller-requested precision.

### L.1 Division refinement

- Replace \`operator/\` with a generator that performs Newton iteration on the reciprocal:
  - Depth 0: \`c0 = a.at(0) / b.at(0)\` (unchanged for backward compatibility)
  - Depth k > 0: refine using \`r_k = r_{k-1} + r_{k-1} * (1 - b * r_{k-1})\` for the reciprocal, then multiply by \`a\` and walk
- Each Newton step doubles the precision; depth k delivers ~ 2^k * 53 bits asymptotically
- Care with \`b\` zero / near-zero: same NaN/inf handling as today

### L.2 sqrt refinement

- Replace \`sqrt\` with a generator that performs Newton iteration:
  - Depth 0: \`c0 = std::sqrt(leading)\` (unchanged)
  - Depth k > 0: refine using \`x_k = (x_{k-1} + a / x_{k-1}) / 2\` walked at increasing precision
- Same care with negative-input and NaN handling as today

## Acceptance

- After this phase lands, the \"elreal \`/\` at depth 0 only\" caveat throughout the docs goes away
- Oracle precision ceiling improves (Phase G helper sees deeper bits for free)
- The Phase I benchmark gets a new row showing \`elreal /\` at depth 1, 2, 3 -- the cost should scale roughly geometrically with depth (each Newton step is a depth-1 op of \`+\` or \`*\`)
- Existing dd / qd / cascade math tests continue to pass when validated against the deeper elreal oracle
- Builds clean under gcc and clang

## Out of scope

- Math function depth (sqrt aside). The transcendentals are Phase M.
- Constants beyond 4 components. That is Phase O.

## Related

- Today's depth-0 cap is documented in \`docs/number-systems/elreal.md\` and \`docs/algorithmic-details/lazy-real-arithmetic.md\`
- The lazy-stream infrastructure already supports arbitrary depth; the work here is per-function generators

Ravenwater:
[...]

Issue #905, opened 2026-05-21
elreal Phase K: Allocator hot-path optimisation -- close the elreal-vs-ereal gap


Ravenwater:
Part of #903. Closes the per-operation throughput gap identified in Phase I (#902).

## Scope

The Phase I baseline (\`docs/algorithmic-details/elreal-performance-baseline.md\`) identified that each elreal arithmetic operator pays for three allocations:

1. A \`std::vector<double>\` for \`_components\` storage
2. A \`std::function<double(std::size_t)>\` for the generator capture (heap-allocates when the capture exceeds the small-buffer threshold -- it does, given the operand copies)
3. Two \`elreal\` copy constructions into the lambda capture (vector walk + generator-pointer copy each)

Per-op cost lands at tens to hundreds of nanoseconds, which is 1.2-3x slower than \`ereal<2>\` on \`+ - *\` at matched precision. This phase addresses the allocations in payoff order.

## Deliverables

Implement in order; each can be a separate PR within this phase:

### K.1 Small-buffer optimisation on \`_components\`

Replace \`std::vector<double>\` storage with a \`small_vector<double, 4>\` (or equivalent \`std::array<double, K> + size_\`). The common case is 1-2 components; SBO eliminates one allocation per op for that case. **This is the single largest item on the list.**

### K.2 Generator type erasure that avoids heap

Replace \`std::function<double(std::size_t)>\` with a tagged union (or intrusive list) of the known generator shapes:

- Depth-1 EFT residual (\`sum_err\`-style, captures two elreals + one double)
- Depth-1 derivative correction (captures one elreal + a derivative value)
- Constant (no capture)
- Degenerate / all-zero (no capture)

Each shape is a fixed-size POD; the tagged union goes on the stack. This eliminates the second per-op allocation.

### K.3 Reference-counted operand sharing

The lambda captures *copies* of both inputs. Switching to \`std::shared_ptr<const Components>\` (or an equivalent intrusive refcount) lets multiple results share an ancestor without copying the component vector. Becomes more valuable once Phase L's depth-2+ generators chain back to ancestors.

### K.4 Batched SIMD on \`two_sum\` and \`two_prod\`

Once K.1-K.3 land, the EFT primitives become a non-trivial fraction of the loop. A batch interface (4-8 EFTs in one SIMD pass) would help reductions and dot products. Only meaningful after the allocator hot path is closed.

## Acceptance

[...]

Issue #903, opened 2026-05-21
Epic: elreal follow-up work -- depth-2+ refinement, range reduction, perf, oracle sweep


Ravenwater:
Follow-up epic to #873 (which shipped phases A through I in 2026). All deliverables of the original epic landed; this epic captures the four work items that were explicitly deferred plus the natural extension of the Phase G validation oracle.

## Background

The original elreal epic (#873) shipped the lazy exact real arithmetic type in nine phases (A-I). Phase H documented the API and algorithmic details; Phase I established the throughput baseline. The baseline measurements and the in-source design comments identified four classes of follow-up work that were out of scope for the initial epic:

1. **Depth-2+ refinement.** Arithmetic operators (`/`, in particular) and math functions (`sqrt`, transcendentals) currently refine to depth 1 only. Deeper refinement requires Newton iteration for division and sqrt, plus higher-precision internal algorithms for the rest. The lazy-stream infrastructure already supports arbitrary depth; only the per-function generators need extending.
2. **Payne-Hanek range reduction for forward trig.** `sin`/`cos`/`tan` use the C++ standard library for depth 0, so large-magnitude arguments lose precision. The path forward is documented in `elreal_impl.hpp` and would compose with a lazy pi-pull from extended-precision constants.
3. **Constants beyond 4 components.** `elreal_pi`, `elreal_e`, and friends reuse the precomputed 4-component expansions from `qd_constants.hpp` (~ 212 bits). Going deeper requires BBP-style on-demand bit extraction for pi, lazy Taylor truncation for e/ln2/ln10, and lazy iteration for sqrt2/sqrt3.
4. **Allocator hot-path optimisation.** Phase I identified the allocator as the dominant per-op cost: each operator allocates a `std::vector` (component storage) and a `std::function` (generator capture) on the heap, plus copies both inputs into the capture. The full breakdown and the four candidate optimisations (small-buffer `_components`, type-erased generator that avoids heap, reference-counted operand sharing, batched SIMD on the EFTs) are documented in `docs/algorithmic-details/elreal-performance-baseline.md` section \"Allocation hot path -- candidates for tuning\".

Plus a natural extension of the Phase G validation oracle:

[...]

PR #912, opened 2026-05-21
perf(elreal): Phase K.1 -- small-buffer optimisation on _components

Ravenwater:
## Summary

Phase K.1 of follow-up epic #903. First sub-phase of Phase K (#905): allocator hot-path optimisation. The Phase I baseline identified per-operator vector allocation as the dominant cost; this PR closes that part of the gap.

K is a multi-PR phase. K.1 is the first and largest payoff item per the Phase I baseline doc: replace \`std::vector<double>\` storage with a small-buffer-optimised type so the common case (depth 1-4 components) pays no heap allocation.

## What landed

- **\`include/sw/universal/number/elreal/lazy_component_buffer.hpp\`** -- new small-buffer storage type: 4-double inline array + spill via \`std::vector\`. 64 bytes total (one cache line). API surface is the minimum elreal needs: \`push_back\`, \`operator[]\`, \`size\`, \`clear\`, \`reserve\`.
- **\`include/sw/universal/number/elreal/elreal_impl.hpp\`** -- migrated \`_components\` from \`std::vector<double>\` to \`lazy_component_buffer\`. The \`components()\` accessor now returns \`const lazy_component_buffer&\`; the single external user in \`manipulators.hpp\` uses \`.size()\` and \`operator[]\` which are unchanged.
- **Two range-for loops** rewritten as indexed loops (the buffer does not provide iterators -- intentional, see the design header).
- **\`docs/algorithmic-details/elreal-performance-baseline.md\`** -- new headline-numbers table for the post-K.1 measurements. Phase I baseline numbers retained for the before/after comparison.
- **\`docs/algorithmic-details/multi-component-arithmetic.md\`** -- section 7.1 picker rule updated: multiplication now favors elreal over \`ereal<2>\` at matched precision (was 1.2x ereal-favored, now 1.9x elreal-favored).

## Headline numbers (gcc 13.3 on 12th Gen Intel i7-12700K)

| Op              | Phase I | Phase K.1 | Speedup |
|-----------------|--------:|----------:|--------:|
| elreal + d1     | 9 Mops/s  | 12 Mops/s  | **1.3x** |
| elreal - d1     | 9 Mops/s  | 14 Mops/s  | **1.6x** |
| elreal * d1     | 8 Mops/s  | 19 Mops/s  | **2.4x** |
| elreal / d0     | 36 Mops/s | ~1 Gops/s  | dominated by inlining once heap alloc is gone |
| elreal sqrt d1  | 14 Mops/s | 30 Mops/s  | **2.1x** |
| elreal exp  d1  | 14 Mops/s | 31 Mops/s  | **2.2x** |
| elreal log  d1  | 14 Mops/s | 24 Mops/s  | **1.7x** |

clang 18.1 shows even larger gains -- the Phase I clang-vs-gcc gap on multiplication (4 vs 8 Mops/s) is closed; both compilers now land at ~ 20 Mops/s.

### Crossover with \`ereal<2>\` at matched precision

- **Addition and subtraction**: ereal<2> still wins by 1.4-2x (was 2-2.7x). The remaining gap is the \`std::function\` generator capture, which Phase K.2 targets.
- **Multiplication**: **elreal now beats ereal<2> by 1.9x** (was 1.2x ereal-favored). \`ereal<N>\` multiplication is O(N) in the eager expansion product; \`elreal *\` is essentially a single \`two_prod\` plus the now-inline result envelope.
- **Division**: not apples-to-apples (elreal depth-0 only until Phase L #906).
- **sqrt / exp / log / pow / trig / hyperbolic**: elreal-only (ereal<N> has no math functions today).

## Test plan

- [x] Builds clean under gcc 13.3 (\`build_elreal/\`)
- [x] Builds clean under clang 18.1 (\`build_clang_elreal/\`)
- [x] All 30 elreal regression tests PASS under both compilers (api, conversion, logic, arithmetic, math, geometry, oracle)
- [x] Phase J oracle sweep across dd / qd / dd_cascade / td_cascade / qd_cascade / ereal<N> PASS under both compilers
- [x] \`benchmark_elreal_performance\` shows the documented gains under both compilers
- [x] CI fast tier green (gcc + clang lite)
- [x] CodeRabbit review

## What this PR does NOT do

- Eliminate the \`std::function\` generator capture. That's **Phase K.2** -- the natural next target now that K.1 has shrunk the component allocation.
- Eliminate the operand copies into the lambda capture. That's **Phase K.3** (refcounted operand sharing).
- Add SIMD/FMA batching to the EFTs. That's **Phase K.4**.

Part of #905 (Phase K of follow-up epic #903).

## Summary by CodeRabbit

* **Documentation**
  * Updated performance baseline with Phase K.1 post-migration results, new post-K.1 tables, per-operator throughput comparisons, and clarified winner profiles and notes on compiler behaviour
  * Revised arithmetic benchmark docs: new throughput table and narrative showing multiplication now favors elreal at matched precision while addition/subtraction still favor ereal<2>

* **Refactor**
  * Optimized component storage with a small-buffer optimization, improving throughput in key cases

[Review Change Stack](https://app.coderabbit.ai/change-stack/stillwater-sc/universal/pull/912?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/algorithmic-details/elreal-performance-baseline.md
+++ b/docs/algorithmic-details/elreal-performance-baseline.md
@@ -1,10 +1,18 @@
 # elreal Performance Baseline
 
-Phase I of epic #873. This document is a baseline measurement -- not a
-performance target. The numbers below come from a single workstation and
-are intended to identify the cost shape of the shipped implementation, so
-that future optimisation work has a starting point and a way to measure
-progress.
+Phase I of epic #873 (baseline) and Phase K.1 of follow-up epic #903
+(small-buffer optimisation on `_components`). This document is a
+baseline measurement -- not a performance target. The numbers below
+come from a single workstation and are intended to identify the cost
+shape of the shipped implementation, so that future optimisation work
+has a starting point and a way to measure progress.
+
+> **Phase K.1 update (#905)**: The `_components` storage migrated from
+> `std::vector<double>` to a small-buffer-optimised
+> `lazy_component_buffer` (inline 4 doubles + spill, see
+> `include/sw/universal/number/elreal/lazy_component_buffer.hpp`).
+> The headline numbers tables below carry both the original Phase I
+> baseline and the post-K.1 measurements.
 
 ## Measurement setup
 
@@ -18,11 +26,12 @@ progress.
   pattern (`a = a + b`) would build up. Per-operation cost is reported
   via `PerformanceRunner` from `include/sw/universal/benchmark/performance_runner.hpp`.
 
-## Headline numbers
+## Headline numbers (Phase I baseline -- pre-K.1)
 
 Throughput in operations per second, rounded. Same workload, two compilers.
 Both elreal and ereal<N> workloads construct fresh operands inside the
-loop body so the per-iteration allocation pattern matches between sides:
+loop body so the per-iteration allocation pattern matches between sides.
+These were the numbers before the K.1 small-buffer optimisation:
 
 | Operation | Budget | gcc 13.3 | clang 18.1 |
 |---|---|---:|---:|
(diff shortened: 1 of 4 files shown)


Ravenwater:
Addressed the single CodeRabbit nitpick in 329c9898: added `assert(i < _size && "lazy_component_buffer index out of range");` to `operator[]` along with `#include <cassert>`. Zero-cost in release (NDEBUG strips it), catches misuse in debug builds.

The invariant `i < _components.size()` already held across all elreal indexing sites (every `_components[i]` call is guarded by a check against `_computed_depth`, which is kept equal to `_components.size()`). The assert just makes it explicit so future contributors get an early signal if they break the invariant.

Validation: all sampled elreal tests (api, arithmetic, math, constants, oracle sweep) PASS under both gcc 13.3 and clang 18.1.

Issue #914, opened 2026-05-21
qd_pi_3 in qd_constants.hpp has wrong value (pi/2 instead of pi/3)


Ravenwater:
## Bug

In \`include/sw/universal/number/qd/math/constants/qd_constants.hpp\`:

\`\`\`cpp
constexpr qd qd_pi_3     (1.570796326794896558, 6.123233995736766036e-17); // pi/3
\`\`\`

The leading component \`1.570796326794896558\` is the value of pi/2, not pi/3 (which is \`1.0471975511965977...\`). Comparison: the line directly below has the same leading value labelled as pi/2:

\`\`\`cpp
constexpr qd qd_pi_2     (1.5707963267948966, 6.123233995736766e-17, ...); // pi/2
constexpr qd qd_pi_3     (1.570796326794896558, 6.123233995736766036e-17); // pi/3   <-- wrong leading value
\`\`\`

The qd_pi_3 entry also only has 2 components instead of the 4 that the other constants in this file carry -- consistent with the entry being a partial / abandoned edit.

## Discovery

Found during the resolution of #911 (high-precision reference string constants for ereal<19>). The 320-digit reference string for pi/3 parses to a leading component of \`1.0471975511965979\`, which mismatches \`qd_pi_3\` and matches the mathematical value of pi/3.

## Fix

Replace the qd_pi_3 line with the correct 4-component expansion of pi/3:

- Component 0: \`1.0471975511965977\` (pi/3 in IEEE 754 double)
- Components 1-3: trailing residuals from the same precomputed expansion source as the other constants in this file

Best derived from the 320-digit reference string \`s_pi_3\` in \`include/sw/math/constants/reference_constants.hpp\` (issue #911), once #913 lands the parse-complexity fix so that parsing the full string into ereal<4> is fast enough.

In the interim, a manual quad-double normalisation of \`pi / 3\` using the Bailey/Hida algorithms would give the correct 4 components.

## Scope

- Single line change in \`qd_constants.hpp\`
- Add a regression test that asserts \`qd_pi_3 + qd_pi_3 + qd_pi_3 == qd_pi\` (within qd tolerance) -- this would have caught the bug

Issue #913, opened 2026-05-21
ereal::parse() complexity blows up on long decimal strings (issue #911 follow-up)


Ravenwater:
## Background

Discovered while resolving #911 (high-precision reference string constants).

\`ereal<N>::parse(const std::string&)\` accumulates digits via Horner's rule:

\`\`\`cpp
ereal<maxlimbs> ten(10.0);
while (pos < str.length()) {
    // ...
    result = result * ten;
    result = result + ereal<maxlimbs>(static_cast<double>(digit));
    // ...
}
\`\`\`

\`result._limb\` grows past \`maxlimbs\` because \`expansion_product\` returns a vector whose size is \`|e| + |f|\` and the assignment overwrites \`_limb\` without truncating. After parsing N digits, \`_limb\` has roughly N+1 entries. Each subsequent multiplication is O(|result| * |ten|) = O(N) per call, with \`linear_expansion_sum\` walking the growing vector. Total complexity: at least O(N^3), possibly worse.

## Symptoms

Measured timings on a 12th Gen Intel i7-12700K (gcc 13.3, \`-O3\`):

| Input length | ereal<2>::parse time | ereal<19>::parse time |
|---:|---:|---:|
| 32 digits | 56 us | 42 us |
| 64 digits | 360 us | 73 us |
| 320 digits | **> 120 s (timeout)** | **> 120 s (timeout)** |

The 320-digit case never completes in any reasonable time, which makes the reference-string constants from #911 unusable through this code path today.

Note that the parsed \`_limb\` size for ereal<2> after a 32-digit string is **28** -- not 2. The parse is not respecting \`maxlimbs\` at all.

## Suggested fixes

Three potential paths, in order of expected payoff:

### 1. Truncate to maxlimbs during the Horner loop

After each \`result = result * ten\` or \`result = result + digit\`, drop trailing limbs once \`result._limb.size() > maxlimbs\`. This keeps the loop's per-step cost bounded at O(maxlimbs^2) rather than growing. Net complexity: O(N * maxlimbs^2) instead of O(N^3). For maxlimbs=19, N=320: from minutes to ~ microseconds.

The truncation is mathematically lossy past maxlimbs but that is the *intended* behaviour for a fixed-precision type. ereal<2> claims 106 bits of precision; carrying 320 limbs through the parse is wasted work.

### 2. Block-Horner: parse digit blocks via double, accumulate via fewer ereal multiplies

[...]

Issue #909, opened 2026-05-21
elreal Phase O: BBP-style and lazy-Taylor constants beyond 4 components


Ravenwater:
Part of #903. Extends the math constants from the 4-component (~ 212-bit) ceiling to arbitrary precision on demand.

## Scope

Phase E.1 (#887, PR #893) shipped \`elreal_pi\`, \`elreal_e\`, \`elreal_ln2\`, \`elreal_ln10\`, \`elreal_sqrt2\`, \`elreal_sqrt3\`, \`elreal_phi\`, plus pi multiples / fractions. All are built from the precomputed 4-component expansions in \`qd_constants.hpp\`, capping the constants at ~ 212 bits. This phase replaces the 4-component fallback with on-demand-deeper computation.

### O.1 BBP-style pi extraction

The Bailey-Borwein-Plouffe formula computes the n-th hexadecimal digit of pi directly, without computing the preceding digits. The lazy generator at depth k pulls 4*k more hex digits (a constant amount of work per depth step). Wire this into \`elreal_pi\` so the existing 4-component leading expansion remains the depth-0 path and the BBP generator kicks in for k >= 4.

### O.2 Taylor truncation for e

The exp(1) series converges fast. Each additional term doubles the precision. Wire as a generator on \`elreal_e\`.

### O.3 Taylor truncation for ln2 / ln10

\`ln(1 + x)\` Taylor on a small-residual argument. The same approach extends to \`ln10 = ln2 + ln(5)\`. Wire as generators.

### O.4 Newton iteration for sqrt2 / sqrt3

Reuse the Phase L Newton-sqrt generator (so this Phase is gated on Phase L).

## Acceptance

- \`elreal_pi.refine_to(1024)\` returns ~ 1024 bits of pi (verified against MPFR or an external high-precision pi source -- consult, do not vendor, the value)
- Same for \`elreal_e\`, \`elreal_ln2\`, \`elreal_ln10\`
- \`elreal_sqrt2\`, \`elreal_sqrt3\` refine to arbitrary depth once Phase L lands
- The picker decision tree note about \"constants are dd/qd-deep only\" goes away
- Builds clean under gcc and clang

## Dependencies

- Phase L (Newton refinement) is required for \`elreal_sqrt2\` / \`elreal_sqrt3\` past 4 components
- Phase N (Payne-Hanek) becomes more useful once \`elreal_pi\` is deeper; argument range for trig grows linearly with pi depth

## Out of scope

[...]

PR #915, opened 2026-05-21
feat(constants): high-precision (~320-digit) reference string constants

### LINKED ISSUE #911: High-precision (~310 decimal digit) string constants for ereal<19> / elreal
## Background

The multi-component types in Universal (\`dd\`, \`qd\`, \`dd_cascade\`, \`td_cascade\`, \`qd_cascade\`, \`ereal<N>\`, \`elreal\`) all share the same precision ceiling derived from IEEE 754 double's exponent range:

- A double has 53 mantissa bits.
- A double's largest finite exponent is ~ 2^1023.
- Multi-component non-overlapping property gives 53 bits per component for values whose leading component is near 1.
- 1023 bits / 53 bits-per-component = **19.3 → at most 19 useful components** for values near 1.
- 1023 bits / 3.32 bits-per-decimal-digit = **~ 310 decimal digits**.

So for any irrational constant whose multi-component expansion saturates the IEEE 754 binade range, a **310-decimal-digit reference string** is sufficient to fill all 19 components of an \`ereal<19>\` and is also the natural source for an \`elreal\` constant generator that delivers arbitrary depth up to that ceiling.

Today's situation:

| Precision tier | Where the constants live |
|---|---|
| double (~ 19 digits) | \`include/sw/math/constants/double_constants.hpp\` |
| long double (~ 19-34 digits) | \`include/sw/math/constants/longdouble_constants.hpp\` |
| qd-equivalent (4 doubles, ~ 64 digits) | \`include/sw/universal/number/qd/math/constants/qd_constants.hpp\` |
| dd-equivalent (2 doubles, ~ 32 digits) | \`include/sw/universal/number/dd/math/constants/dd_constants.hpp\` |
| elreal (reuses the 4-component qd expansions) | \`include/sw/universal/number/elreal/math/constants/elreal_constants.hpp\` |
| **~ 310-digit ultra-precision strings** | **missing -- this issue** |

\`ereal<N>\` already has a working \`parse(const std::string&)\` (\`include/sw/universal/number/ereal/ereal_impl.hpp:314\`) that converts a decimal string into the multi-component form. \`elreal\` has the same via its string constructor. So **once we have the strings, the multi-component initialisation is a one-liner**: \`ereal<19> pi(\"3.14159265358979...\");\`.

The missing piece is the strings themselves.

## Scope

Provide a header that holds 310-decimal-digit reference strings for the basic math constants used across the multi-component types, plus a small set of factory functions that materialise them as \`ereal<N>\` / \`elreal\` values on demand.

### 1. Reference-string header

New header \`include/sw/math/constants/string_constants.hpp\` (or a name closer to the precision-class naming convention -- naming TBD; see \"Open questions\" below) holding string constants for at least:

- Pi: \`pi\`, \`pi/2\`, \`pi/3\`, \`pi/4\`, \`2*pi\`, \`1/pi\`, \`2/pi\`
- Euler's e: \`e\`, \`1/e\`
- Golden ratio: \`phi\`, \`1/phi\`
- Square roots: \`sqrt(2)\`, \`sqrt(3)\`, \`sqrt(5)\`, \`1/sqrt(2)\`
- Logarithms: \`ln(2)\`, \`ln(10)\`, \`log2(e)\`, \`log2(10)\`, \`log10(e)\`, \`log10(2)\`
- Trig anchors: \`2/sqrt(pi)\` (the erf scaling)
- Euler-Mascheroni gamma (only the ones already in \`longdouble_constants.hpp\`; do not expand the set beyond what is already used downstream)

Each string is at least 310 significant decimal digits, ASCII-only, single-line per constant, with a comment naming the constant and its source (so future contributors can re-verify).

### 2. Materialisation helpers

A small set of factory functions in the same header or a sibling:

\`\`\`cpp
template<unsigned N> ereal<N> make_pi()    { ereal<N> v; v.parse(s_pi);    return v; }
template<unsigned N> ereal<N> make_e()     { ereal<N> v; v.parse(s_e);     return v; }
// ... etc
elreal make_pi_elreal()  { return elreal(s_pi);    /* string ctor */ }
elreal make_e_elreal()   { return elreal(s_e);     }
// ... etc
\`\`\`

Plus convenience aliases for the most-used widths (\`ereal<19>\`, the current natural ceiling).

### 3. Cross-validation tests

A test target at \`static/math/constants/string_constants.cpp\` (or similar) that:

- Parses each 310-digit string into \`ereal<19>\`
- Verifies the leading components match the existing \`qd_constants.hpp\` 4-component expansions (sanity)
- Cross-validates against the elreal oracle (Phase J machinery from #904)
- Reports the achieved digit count per constant (should be >= ~ 305 once normalisation kicks in)

## Research questions to settle in this issue

Before writing the strings, decide:

### Q1: Generation source

Three options for producing the strings. Pick one (or a mix):

- **External tool, vendored output.** Generate offline with \`mpmath\` (Python), PARI/GP, or Wolfram. Document the exact command used to regenerate. Lowest risk; reproducibility recipe lives next to the strings.
- **Public-domain reference tables.** Pi, e, ln(2), sqrt(2) are widely tabulated to thousands of digits in public-domain sources (NIST, Wolfram MathWorld, Y-cruncher logs). Copy the first 310 digits verbatim. Easy but error-prone (transcription).
- **Universal's own \`efloat\`.** Generate at runtime using arbitrary-precision computation in Universal itself. Self-hosting and self-verifying, but slow and adds a build-time dependency on \`efloat\`'s math suite (which may not exist yet for all the constants we need).

**Recommendation**: option 1 (mpmath), with the generation script stored in \`tools/scripts/generate_string_constants.py\` so the strings can be re-derived deterministically.

### Q2: Storage form

- Single \`std::string_view\` per constant in a header (the simplest)?
- \`constexpr const char* const s_pi = \"3.14159...\"\` (works in C++17+)?
- A \`std::array<char, 320>\` per constant (avoids the pointer indirection)?

**Recommendation**: \`constexpr inline const std::string_view\` per constant -- works in C++17, no static-init order issues, and the parse helper takes a string view natively.

### Q3: Verification

How do we know the 310 digits are correct?

- Cross-check option A: regenerate from a second source (e.g. mpmath *and* PARI/GP) and verify byte-for-byte equality.
- Cross-check option B: parse into \`ereal<19>\` and verify the first 64 digits (the \`qd_constants.hpp\` width) match the existing qd expansions component-by-component.
- Cross-check option C: parse into \`elreal\` and run the Phase J oracle helper against the existing dd / qd / cascade constants.

**Recommendation**: do all three. A and B at table-generation time; C as a one-shot regression test.

### Q4: Naming and location

The existing \`include/sw/math/constants/\` directory has \`double_constants.hpp\` and \`longdouble_constants.hpp\` -- the naming reflects the *storage type*. A 310-digit string is not tied to a storage type; it is a *reference value*.

Options:
- \`include/sw/math/constants/string_constants.hpp\` -- matches the storage form
- \`include/sw/math/constants/reference_constants.hpp\` -- matches the semantic role
- \`include/sw/math/constants/decimal_constants.hpp\` -- matches what the file contains

**Recommendation**: \`reference_constants.hpp\` to leave room for non-decimal reference forms in the future (e.g. hex-digit pi, byte-array pi for BBP-style extraction in Phase O #909).

## Acceptance

- A header lands at the agreed location with 310-digit-or-better string constants for the math-constants set above
- The strings are reproducible from a script committed under \`tools/scripts/\`
- The factory functions materialise each constant into \`ereal<N>\` and \`elreal\` correctly
- The test target verifies cross-consistency with the existing 4-component \`qd_constants.hpp\` expansions
- The Phase J oracle sweep (#904) catches any disagreement
- Documentation update in \`docs/algorithmic-details/\` explaining the precision ceiling (1023 bits / 19 components / 310 digits) and where the strings come from

## Out of scope

- Generating the strings *at compile time* via constexpr arithmetic. Too complex for the gain; we have working runtime parsers.
- Replacing the existing 4-component \`qd_constants.hpp\` expansions. Those stay -- they are the depth-0 / fast path. The new strings serve callers who need \`ereal<N>\` for N > 4 or \`elreal\` constants beyond the 4-component fallback.
- Constants beyond the basic set listed above. The criterion is: only constants that are already exposed somewhere in Universal's math API today.

## Related

- Phase O of the elreal follow-up epic (#909) -- BBP-style and lazy-Taylor *generators*. The work here is the complementary precomputed-string approach. The two are not mutually exclusive: precomputed strings serve callers who want a value materialised once; BBP / Taylor generators serve callers who want arbitrary depth lazily.
- Phase J oracle sweep (#904) -- becomes the natural validator for these strings.
- Existing constants infrastructure: \`include/sw/math/constants/\`, \`include/sw/universal/number/qd/math/constants/\`, \`include/sw/universal/number/elreal/math/constants/\`
- Existing parse machinery: \`ereal<N>::parse()\` in \`include/sw/universal/number/ereal/ereal_impl.hpp\`; \`elreal\` string constructor in \`include/sw/universal/number/elreal/elreal_impl.hpp\`

## Notes

The 310-digit ceiling is specific to **values near 1**. For constants whose magnitude differs from 1, the achievable component count shifts:

- \`2*pi\` (~ 6.28): leading binade is \`[4, 8)\`, so the per-component bit budget shifts by 2 bits. 19 components still gives ~ 309 digits.
- \`1/pi\` (~ 0.318): leading binade is \`[1/4, 1/2)\`, so the budget shifts by 2 bits the other way. Same ~ 309-digit reach.

The 310-digit number is a practical upper bound that covers any constant in the multi-component range with margin. No need to size-specialise per constant.

Ravenwater:
## Summary

Resolves #911. Adds decimal-string reference constants at ~320 digits of precision -- the natural ceiling for any multi-component type built on double components (floor(1023/53)=19 components for values near 1, ~ 1023/3.32=308 decimal digits, with margin).

## What landed

- **\`tools/scripts/generate_reference_constants.py\`** -- mpmath-based generation script at 350 internal digits / 320 emitted. Pure-Python reproducibility recipe; the script is the canonical source of truth for the strings in the header.
- **\`include/sw/math/constants/reference_constants.hpp\`** -- 24 string constants as \`inline constexpr std::string_view\`, each 320 decimal digits, plus template factory helpers \`make_pi<T>()\`, \`make_e<T>()\`, etc. that materialise any type T satisfying \`T(std::string)\` (covers \`ereal<N>\` and \`elreal\`).
- **\`elastic/ereal/math/constants/reference_constants.cpp\`** -- cross-validation test against the precomputed 4-component qd expansions in qd_constants.hpp. Target: \`er_const_reference_constants\`.

## Constants set

- pi family: \`pi\`, \`pi/2\`, \`pi/3\`, \`pi/4\`, \`2*pi\`, \`3*pi\`, \`1/pi\`, \`2/pi\`
- e family: \`e\`, \`1/e\`
- phi family: \`phi\`, \`1/phi\`
- square roots: \`sqrt(2)\`, \`sqrt(3)\`, \`sqrt(5)\`, \`1/sqrt(2)\`
- logarithms: \`ln(2)\`, \`ln(10)\`, \`log2(e)\`, \`log2(10)\`, \`log10(e)\`, \`log10(2)\`
- erf scaling: \`2/sqrt(pi)\`
- Euler-Mascheroni gamma

## Research questions from the issue body

The issue body posed four design questions; the answers chosen here:

1. **Generation source**: **mpmath** via a committed Python script in \`tools/scripts/\`. Reproducible from a single command.
2. **Storage form**: **\`inline constexpr std::string_view\`** -- works in C++17, no static-init order issues, cheap to copy. Factory helpers construct \`std::string\` only at the parse call site.
3. **Verification**: **cross-check against qd_constants.hpp** -- the leading 4 components of each parsed string must match the precomputed qd expansion. (Cross-check against an independent generation source like PARI/GP is desirable as a future hardening step; deferred.)
4. **Naming and location**: **\`reference_constants.hpp\`** in \`include/sw/math/constants/\` (next to \`double_constants.hpp\` and \`longdouble_constants.hpp\`). The name names the semantic role rather than the storage type.

## Known limitations

Two issues were filed during this PR for follow-up:

- **#913**: \`ereal::parse()\` has quadratic-or-worse complexity on long strings. The full 320-digit reference strings exist in the header and are mathematically correct, but the cross-validation test uses 80-character truncations to stay under the parse-perf cliff. The full strings will become practically usable once #913 lands the parse complexity fix (truncate \`_limb\` to \`maxlimbs\` during the Horner accumulation).
- **#914**: \`qd_pi_3\` in \`qd_constants.hpp\` holds the wrong value (~ pi/2 instead of pi/3) and only 2 components. The cross-check test skips \`pi_3\` specifically and points at #914 in a comment. All other constants cross-check cleanly.

## Test plan

- [x] Builds clean under gcc 13.3 and clang 18.1 (-O3)
- [x] \`er_const_reference_constants\` PASS under both compilers
- [x] Cross-validation against \`qd_constants.hpp\` confirms leading components agree exactly and trailing components agree within 1e-30 relative tolerance for all constants except \`qd_pi_3\` (skipped, see #914)
- [ ] CI fast tier green (gcc + clang lite)
- [ ] CodeRabbit review

## Out of scope (deferred to follow-up)

- Cross-source verification (re-generate via PARI/GP and verify byte-for-byte) -- listed as Q3 option A in the issue body; the qd-cross-check is the immediate verifier
- Fixing the ereal::parse perf bomb -- #913
- Fixing qd_pi_3 -- #914
- Adding the strings to elreal's constant set so it can refine past 4 components -- depends on Phase O (#909) and on #913

Resolves #911.

## Summary by CodeRabbit

* **New Features**
  * High-precision reference constants for many mathematical values (π, e, φ, square roots, logarithms, Euler–Mascheroni, etc.).
  * Factory functions to materialize these references into custom numeric types.

* **Tests**
  * Added a validation test suite that checks materializations and performs sanity comparisons.

* **Tools**
  * New script to regenerate the high-precision reference strings.

[Review Change Stack](https://app.coderabbit.ai/change-stack/stillwater-sc/universal/pull/915?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Ravenwater:
Both CodeRabbit nitpicks addressed in d2633289:

1. **Test coverage**: Added cross-checks for the 5 missing constants that have qd counterparts (`log2e`, `log2_10`, `log10e`, `log10_2`, `two_over_sqrt_pi`). The qd names are unusual short-form (`qd_lge`, `qd_lg10`, `qd_loge`, `qd_log2`, `qd_2_sqrtpi`); the test maps each reference string to the correct qd constant and documents the naming mismatch inline. Also added as-double sanity checks for `three_pi` and `euler_gamma`, which have no qd counterpart at all.

2. **`from_reference<T>` constraint**: Added `static_assert(std::is_constructible_v<T, std::string>, ...)` per the proposed patch. Gives a clear compile-time diagnostic at the API boundary instead of a deep template error inside T's constructor.

Validation: all 22 cross-checks PASS under gcc 13.3 and clang 18.1 (17 qd-cross-checks + sanity checks for three_pi, euler_gamma, pi, e + skip for pi_3 per #914).

Issue #923, opened 2026-05-22
Epic: McCleeary LFPERA implementation of elreal


Ravenwater:
## Phase tracker

- [x] Phase 1 (#925): `block<FpType>` representation
- [x] Phase 2 (#926): `ZBCL<FpType>` co-list
- [x] Phase 3 (#927): twoSum / twoMult / twoDiv on blocks
- [x] Phase 4 (#928): threeAdd + ZBCL addition
- [x] Phase 5 (#929): infinite summation
- [x] Phase 6 (#930): negation, multiplication, division
- [x] Phase 7 (#931): transcendental functions and numbers
- [x] Phase 8 (#932): real floating-point conversion
- [x] Phase 9 (#933): evaluation against qd + hardware block-shape study

## Goal

Reimplement \`elreal\` faithfully following Chapter 4 of McCleeary's 2019 dissertation, \"Lazy Floating Point Exact Real Arithmetic\" (Univ. of Iowa).

The prior implementation (PRs #883-#900, backed out in PR #922) built Shewchuk-style multi-component storage with McCleeary-style lazy materialisation grafted on. The mismatch broke the non-overlap property at refinement depth 3+ (see #921). The Shewchuk invariant \`|c_{i+1}| <= ulp(c_i)/2\` requires a renormalisation sweep that the lazy left-to-right framework cannot perform.

McCleeary's algorithm avoids this entirely through the 0-overlap gap (the \"imp\" bit between consecutive blocks), which absorbs all downstream carry-ups without any renormalisation pass.

### Block parameterisation: \`block<FpType>\` -- host FP type, not bare \`k\`

The block is templated on the host floating-point type, NOT on a bare significand-bit count \`k\`. This binds the block to a concrete hardware target and carries both the significand width (the dissertation's \`k = digits - 1\`) AND the exponent width in one parameter.

\`\`\`cpp
// Standard hardware targets
using elreal_h    = elreal<half>;       // k=10, e=5   -- FP16 GPU tensor cores
using elreal_bf   = elreal<bfloat16>;   // k=7,  e=8   -- BF16 TPU / GPU
using elreal_s    = elreal<float>;      // k=23, e=8   -- FP32 CPU/GPU baseline
using elreal_d    = elreal<double>;     // k=52, e=11  -- FP64 CPU baseline
using elreal_q    = elreal<quad>;       // k=112, e=15 -- FP128 (__float128)

// Custom hardware targets (FPGA, accelerator research)
[...]

PR #1015, opened 2026-05-27
feat(bfloat16): add ldexp/frexp/scalbn/logb/ilogb exponent functions (#941)

### LINKED ISSUE #941: Add ldexp / frexp / scalbn / logb / ilogb to bfloat16 math library
## Summary

`bfloat16` is missing the floating-point exponent-manipulation functions that the C++ `<cmath>` family provides for native types: `std::ldexp`, `std::frexp`, `std::scalbn`, `std::logb`, `std::ilogb`. The peer types (`cfloat<>`) expose these; `bfloat16` does not.

## What's missing

`include/sw/universal/number/bfloat16/math/functions/exponent.hpp` currently covers `exp`, `exp2`, `exp10`, `expm1`. It does NOT cover:

| Function | Purpose |
|---|---|
| `ldexp(x, n)` | x * 2^n |
| `frexp(x, &exp)` | split x into mantissa in [0.5, 1) and an integer exponent |
| `scalbn(x, n)` | same as ldexp for IEEE radix |
| `logb(x)` | unbiased exponent as a float |
| `ilogb(x)` | unbiased exponent as an int |

The cfloat counterpart at `include/sw/universal/number/cfloat/cfloat_impl.hpp:4355` shows the pattern:

```cpp
template<...> constexpr cfloat<...> frexp(const cfloat<...>& x, int* exp) {
    *exp = x.scale();
    cfloat<...> fraction(x);
    fraction.setexponent(0);
    return fraction;
}
template<...> constexpr cfloat<...> ldexp(const cfloat<...>& x, int exp) {
    cfloat<...> result(x);
    int xexp = x.scale();
    result.setexponent(xexp + exp);  // TODO: doesn't work for subnormals
    return result;
}
```

## How it surfaced

Discovered while writing Phase 4 of the McCleeary `elreal` work (PR #940 / epic #923). The `twoSumRN` primitive in `threeAdd.hpp` aligns two blocks to a common exponent via `ldexp(v, delta)`, which works for `float`, `double`, and `cfloat<>` (the latter via ADL). `bfloat16` falls through to a fallback that converts through `double` -- functionally correct, but losing the "use host arithmetic directly" guarantee.

PR #940 sidestepped this by restricting Phase 4 tests to `{float, double}`. Once this issue is fixed, the elreal Phase 4+ tests can include `bfloat16` in the FpType sweep.

## Suggested implementation

Two paths, both reasonable:

1. **Delegate to double** (matches the existing `exp`/`log`/`sin` style in bfloat16):
   ```cpp
   inline bfloat16 ldexp(bfloat16 x, int n) {
       return bfloat16(std::ldexp(double(x), n));
   }
   inline bfloat16 frexp(bfloat16 x, int* exp) {
       double f = std::frexp(double(x), exp);
       return bfloat16(f);
   }
   ```
   Pros: minimal code, hardware FP cost. Cons: round-trip through double; subnormals could trigger double rounding.

2. **Bit manipulation** (matches the cfloat pattern):
   ```cpp
   inline bfloat16 ldexp(bfloat16 x, int n) {
       bfloat16 result(x);
       result.setexponent(x.scale() + n);  // uses bfloat16's existing scale() + setexponent
       return result;
   }
   ```
   Pros: stays in bfloat16. Cons: needs bfloat16 to expose `setexponent()` (currently only `setbits()`); subnormal handling needs care.

Option 1 is the lower-risk first cut; option 2 can land later as an optimisation if the bit-manipulation primitives gain test coverage.

## Where to land

- Add to `include/sw/universal/number/bfloat16/math/functions/exponent.hpp` (already exists, currently only covers `exp` family).
- Add tests under `static/bfloat16/math/` (mirror the structure used for cfloat).

## Discovery date / context

2026-05-23, during elreal Phase 4 (PR #940). The PR has a `TODO` comment in `threeAdd.hpp::detail_mccleeary::ldexp_block` noting the bfloat16 fallback path; once this issue ships, that comment can be removed and the bfloat16 sweep added to Phase 4 tests.

Ravenwater:
## Summary
`bfloat16` was missing the `<cmath>` exponent-manipulation family that `cfloat<>` exposes. Added **`ldexp`, `frexp`, `scalbn`, `logb`, `ilogb`** to `bfloat16/math/functions/exponent.hpp`, delegating to the host `double` routines (Option 1 in the issue) -- matching bfloat16's existing `exp`/`log`/`sin` style. A bfloat16 is exactly representable in `double`, so the only rounding is the final result back to bfloat16's 8-bit significand.

`exponent.hpp` is already included by `bfloat16/mathlib.hpp`, so the functions are available via ADL wherever the mathlib is in scope (e.g. the elreal `threeAdd` `twoSumRN` alignment path that motivated this).

## Changes
- `include/sw/universal/number/bfloat16/math/functions/exponent.hpp` -- 5 new functions.
- `static/float/bfloat16/math/exponent.cpp` -- property-based regression test.

## Test (semantic, not implementation-derived)
Scaling by a power of two is exact in bfloat16, so the test asserts exact identities: `frexp`/`ldexp` round-trip, `|fraction| in [0.5,1)`, `scalbn==ldexp`, `ilogb==frexp_exp-1`, `logb==ilogb`, and `2^ilogb(x) <= |x| < 2^(ilogb(x)+1)`. Verified it catches a broken `ldexp` (exit non-zero).

| Target | gcc | clang |
|--------|-----|-------|
| bfloat16 exponent functions | PASS | PASS |
| bfloat16 mathlib (umbrella, no conflicts) | PASS | PASS |

## Follow-up (out of scope; elreal epic #923/#940)
Drop the bfloat16 fallback TODO in `threeAdd.hpp::detail_mccleeary::ldexp_block` and add bfloat16 to the Phase 4 FpType sweep.

## Test plan
- [ ] Fast CI passes
- [ ] Promote to ready: `gh pr ready`

Resolves #941

## Summary by CodeRabbit

* **New Features**
  * Added exponent manipulation functions for bfloat16: `frexp`, `ldexp`, `scalbn`, `logb`, and `ilogb` for enhanced mathematical operations.

* **Tests**
  * Added comprehensive test suite validating exponent manipulation functions with multiple mathematical properties and edge cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/stillwater-sc/universal/pull/1015?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/include/sw/universal/number/bfloat16/math/functions/exponent.hpp
+++ b/include/sw/universal/number/bfloat16/math/functions/exponent.hpp
@@ -28,4 +28,35 @@ inline bfloat16 expm1(bfloat16 x) {
 	return bfloat16(std::expm1(double(x)));
 }
 
+// ---- exponent manipulation (issue #941) ----
+// bfloat16 delegates to the host double routines (matching the exp/log family
+// above); a bfloat16 value is exactly representable in double, so the only loss
+// is the final round of the result back to bfloat16's 8-bit significand.
+
+// Decompose x into a normalized fraction in [0.5, 1) and an integer power of two:
+// x == frexp(x, &e) * 2^e. Writes the exponent to *exp.
+inline bfloat16 frexp(bfloat16 x, int* exp) {
+	return bfloat16(std::frexp(double(x), exp));
+}
+
+// Multiply x by 2^n.
+inline bfloat16 ldexp(bfloat16 x, int n) {
+	return bfloat16(std::ldexp(double(x), n));
+}
+
+// Multiply x by FLT_RADIX^n; FLT_RADIX == 2 here, so identical to ldexp.
+inline bfloat16 scalbn(bfloat16 x, int n) {
+	return bfloat16(std::scalbn(double(x), n));
+}
+
+// Unbiased radix-2 exponent of x, as a floating-point value.
+inline bfloat16 logb(bfloat16 x) {
+	return bfloat16(std::logb(double(x)));
+}
+
+// Unbiased radix-2 exponent of x, as an int.
+inline int ilogb(bfloat16 x) {
+	return std::ilogb(double(x));
+}
+
 }} // namespace sw::universal
--- a/static/float/bfloat16/math/exponent.cpp
+++ b/static/float/bfloat16/math/exponent.cpp
@@ -1,0 +1,142 @@
+// exponent.cpp: test suite for bfloat16 exponent-manipulation functions
+//               ldexp / frexp / scalbn / logb / ilogb (issue #941)
+//
+// Copyright (C) 2017 Stillwater Supercomputing, Inc.
+// SPDX-License-Identifier: MIT
+//
+// This file is part of the universal numbers project, which is released under an MIT Open Source license.
+#include <universal/utility/directives.hpp>
+#include <cmath>
+#include <universal/number/bfloat16/bfloat16.hpp>
+#include <universal/verification/test_suite.hpp>
+
+namespace sw { namespace universal {
+
+	// Verify the exponent-manipulation functions by their mathematical
+	// properties (not by re-deriving the double delegation): scaling by a power
+	// of two is exact in bfloat16, so these identities must hold exactly.
+	int VerifyExponentFunctions(bool reportTestCases) {
+		int nrOfFailedTestCases = 0;
+
+		// representative finite, normal bfloat16 values (all exactly representable)
+		const float values[] = { 1.0f, 0.5f, 2.0f, 3.5f, 0.75f, 100.0f, 0.015625f, -4.0f, -0.25f };
+		const int   shifts[] = { -5, -1, 0, 1, 3, 7 };
+
+		for (float fv : values) {
+			bfloat16 x(fv);
+
+			// frexp: x == fraction * 2^e, with |fraction| in [0.5, 1)
+			int e = 0;
+			bfloat16 fraction = frexp(x, &e);
+			double fabsd = std::abs(double(fraction));
+			if (!(fabsd >= 0.5 && fabsd < 1.0)) {
+				if (reportTestCases) std::cout << "FAIL frexp fraction out of [0.5,1): " << fv << " -> " << double(fraction) << '\n';
+				++nrOfFailedTestCases;
+			}
+			if (ldexp(fraction, e) != x) {
+				if (reportTestCases) std::cout << "FAIL frexp/ldexp roundtrip: " << fv << '\n';
+				++nrOfFailedTestCases;
+			}
+
+			// ilogb / logb / frexp exponent relationship: frexp exponent == ilogb + 1
+			int il = ilogb(x);
+			if (il != e - 1) {
+				if (reportTestCases) std::cout << "FAIL ilogb vs frexp exponent: " << fv << " ilogb=" << il << " frexp_e=" << e << '\n';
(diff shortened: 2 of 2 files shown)


Ravenwater:
would it not be better to marshal through float instead of double? Float shares the same exponent range, and thus the conversions are cheaper.

include/sw/universal/number/bfloat16/math/functions/exponent.hpp
```diff
@@ -28,4 +28,35 @@ inline bfloat16 expm1(bfloat16 x) {
...
+
+// Decompose x into a normalized fraction in [0.5, 1) and an integer power of two:
+// x == frexp(x, &e) * 2^e. Writes the exponent to *exp.
+inline bfloat16 frexp(bfloat16 x, int* exp) {
+	return bfloat16(std::frexp(double(x), exp));
+}
+
+// Multiply x by 2^n.
+inline bfloat16 ldexp(bfloat16 x, int n) {
+	return bfloat16(std::ldexp(double(x), n));
```
Ravenwater, line 44: would it not be better to marshal through float instead of double? Float shares the same exponent range, and thus the conversions are cheaper.

include/sw/universal/number/bfloat16/math/functions/exponent.hpp
```diff
@@ -28,4 +28,35 @@ inline bfloat16 expm1(bfloat16 x) {
...
+}
+
+// Multiply x by FLT_RADIX^n; FLT_RADIX == 2 here, so identical to ldexp.
+inline bfloat16 scalbn(bfloat16 x, int n) {
+	return bfloat16(std::scalbn(double(x), n));
```
Ravenwater, line 49: would it not be better to marshal through float instead of double? Float shares the same exponent range, and thus the conversions are cheaper.

include/sw/universal/number/bfloat16/math/functions/exponent.hpp
```diff
@@ -28,4 +28,35 @@ inline bfloat16 expm1(bfloat16 x) {
...
+}
+
+// Unbiased radix-2 exponent of x, as a floating-point value.
+inline bfloat16 logb(bfloat16 x) {
+	return bfloat16(std::logb(double(x)));
```
Ravenwater, line 54: would it not be better to marshal through float instead of double? Float shares the same exponent range, and thus the conversions are cheaper.

include/sw/universal/number/bfloat16/math/functions/exponent.hpp
```diff
@@ -28,4 +28,35 @@ inline bfloat16 expm1(bfloat16 x) {
...
+}
+
+// Unbiased radix-2 exponent of x, as an int.
+inline int ilogb(bfloat16 x) {
+	return std::ilogb(double(x));
```
Ravenwater, line 59: would it not be better to marshal through float instead of double? Float shares the same exponent range, and thus the conversions are cheaper.

Ravenwater (commented):
would it not be better to marshal through float instead of double? Float shares the same exponent range, and thus the conversions are cheaper.

Ravenwater:
Addressed in follow-up #1017: the five exponent functions now marshal through `float` instead of `double` (bfloat16 shares float32's 8-bit exponent field, so the conversion is exact and cheaper), and the test gained IEEE special-case coverage (0/inf/NaN) per CodeRabbit. Resolving these threads.

Issue #942, opened 2026-05-23
Add exact-EFT fallback for odd-precision FpType in block_eft two_prod


Ravenwater:
## Summary

The generic Dekker-style `two_prod_host` in `include/sw/universal/number/elreal/block_eft.hpp` is exact only when the host FpType has even precision `p`. For odd-`p` types (e.g. `cfloat<24,5>` with `digits=19`, `half`/`cfloat<16,5>` with `digits=11`, `bfloat16` with `digits=7`), the Veltkamp split requires `2*ceil(p/2) <= p`, which fails for odd `p`, leaving ~ulp^2 residual error in the EFT.

The native `double` specialization already uses FMA via `error_free_ops.hpp` and is exact regardless. Other FpTypes need the same treatment so the McCleeary elreal phase work can include them.

## Background

Dekker's two_prod via Veltkamp split:

```
S = 2^ceil(p/2) + 1
temp = S * a
a_hi = temp - (temp - a)
a_lo = a - a_hi
// similarly split b, then:
r = ((a_hi*b_hi - p) + a_hi*b_lo + a_lo*b_hi) + a_lo*b_lo
```

For the residual `r` to be exact, the partial product `a_hi * b_hi` must fit in `p` bits, which requires `2*s <= p` where `s = ceil(p/2)`. This holds iff `p` is even.

For odd `p`:
- p=19 (cfloat<24,5>): s=10, 2s=20 > 19, off by 1 bit
- p=11 (half): s=6, 2s=12 > 11, off by 1 bit
- p=7 (bfloat16 per Universal): s=4, 2s=8 > 7, off by 1 bit

Result: residual carries ~ulp^2 error instead of being exact.

## Verification

The cfloat arithmetic itself is IEEE-correct -- this is purely an EFT-algorithm property. Diagnostic from #940:

```
cf24 (p=19 odd)   1.3 * 1.7 = 2.21  residual=6.103510e-07  ulp(p)/2=3.814697e-06  ratio=0.16
cf32 (p=24 even)  1.3 * 1.7 = 2.21  residual=-5.722046e-08 ulp(p)/2=1.192093e-07  ratio=0.48
float (p=24)      1.3 * 1.7 = 2.21  residual=-5.722046e-08 ulp(p)/2=1.192093e-07  ratio=0.48
```

All three are within `ulp(p)/2` -- IEEE-correct rounding. The issue surfaces only when we go through the SECOND step (Dekker split) for odd `p`.

## Fix options

[...]

PR #1023, opened 2026-05-27
test(elreal): independent exact dyadic oracle for Phase 1-4 building blocks

### LINKED ISSUE #1022: elreal Phase 5 precondition: port the exact dyadic oracle and re-assert Phases 1-4 bit-exactly
## Why (the ereal lesson, applied to elreal)

The ereal/Priest effort (epic #987, retrospective in `docs/assessments/ereal-retrospective.md`) established a hard rule: **when a result is supposed to be exact, validate it against an INDEPENDENT exact oracle that shares no code with the algorithm under test** -- not against a reference that is merely "more precise," and never against a structural/self-referential check. ereal got there via `include/sw/universal/verification/dyadic_exact.hpp` (exact dyadic rationals backed by `einteger`) used in `elastic/ereal/arithmetic/exact_value_oracle.cpp`:

```
exact_value(a + b) == exact_value(a) + exact_value(b)   // bit-exact dyadic equality
```

The McCleeary elreal rebuild (epic #923) has completed Phases 1-4 -- block<FpType> (#925), ZBCL<FpType> (#926), block-level EFTs (#927), threeAdd + ZBCL add() (#928). The building blocks are validated against McCleeary's dissertation theorems (threeAdd's value-preservation / dominance / exponent-bound / no-shrink; ZBCL 0-overlap; prefix-stability), which is the right *methodology*. **But the value-preservation half rests on `long double`, which is not an exact oracle**, and that is exactly the weakness ereal eliminated.

This issue is the **precondition for Phase 5+ (#929 infinite summation, #930 negation/mul/div, ...)**: lock down a *validated* baseline before stacking more arithmetic on it, so we do not repeat the ereal "compose new ops on unvalidated blocks" trap one layer up.

## The gap (evidence)

1. **No independent exact oracle.** Nothing under `elastic/elreal/` uses dyadic/MPFR/__float128/multiprecision; every value check uses `long double`.
   - On x86_64, `long double` is a 64-bit mantissa (~11 bits over `double`): catches gross errors, **cannot certify bit-exactness** (the exact sum/product of two doubles can need ~107 bits).
   - On MSVC / most ARM / RISC-V, `long double == double`: **zero headroom over the type under test** -- the value checks are near-vacuous there (the `threeAdd.cpp` comment admits this).
2. **Loose tolerance on an exact type.** `elastic/elreal/arithmetic/addition.cpp` allows ~4 FpType-ulp. Representable cases (`1 + 0.25 -> exactly 1.25`) should assert `diff == 0`; a 4-ulp band would not fail on a sub-ulp regression.
3. **Thin inputs.** `addition.cpp` uses ~9 hand-picked, all-exactly-representable operands over only double/float; no catastrophic-cancellation, many-limb, or subnormal stress. (`threeAdd.cpp` does randomize -- keep that.)

## Deliverable

Port the dyadic oracle to elreal and re-assert the Phases 1-4 building blocks bit-exactly.

- [ ] **Exact-value mapping.** Add an `exact_value()` for elreal primitives, sharing no code with the block/EFT/threeAdd algorithms:
  - block: `exact_value(block) = dyadic::from_double(double(b.v))` shifted by `b.exp` (the `<<`/scale ops in `dyadic_exact.hpp`). `double(b.v)` is exact for every elreal FpType (float/half/bfloat16/cfloat<=24 sig bits all widen to double exactly).
  - ZBCL prefix: `exact_value(take(n)) = sum of the block dyadics` (blocks are non-overlapping, so the sum is exact).
- [ ] **Re-assert Phase 3 EFTs** (`two_sum`, `two_mult`, `two_div`) with exact dyadic equality where the EFT is exact: `exact(hi)+exact(lo) == exact(a) {+,*} exact(b)`. (two_div / inexact cases: keep the relative-error bound, as ereal does.) Note the cfloat<24,5> subnormal-residual floor from #942 is a *representability* limit -- assert against what the type can hold, exactly.
- [ ] **Re-assert Phase 4 add()** with exact equality: `exact_value(take(N)) == exact(a) + exact(b)` once `N >= number of emitted blocks`. Drop the 4-ulp tolerance on representable cases to `== 0`.
- [ ] **Randomized + adversarial sweeps** (seeded, deterministic): catastrophic cancellation, widely-separated exponents (many-limb expansions), subnormals, signed zeros -- not just the hand-picked values.
- [ ] **Oracle-headroom guard.** These exact checks must run where the oracle is genuinely independent of the type under test. The oracle is dyadic (arbitrary precision via einteger), so it is exact regardless of `long double` -- but ensure no path silently falls back to a `long double` reference on long-double-aliased platforms.

## Acceptance criteria

- New test (e.g. `elastic/elreal/arithmetic/exact_value_oracle.cpp`) asserts bit-exact dyadic equality for blocks, the three EFTs (exact cases), threeAdd, and add(), over both hand-picked and randomized/adversarial inputs, for every supported FpType (double, float, half, bfloat16, cfloat<24,5>, cfloat<32,8>).
- Existing Phase 1-4 tests that used loose `long double` tolerances on representable cases are tightened to exact comparison (or clearly justified where the *type* cannot represent the exact result, e.g. the #942 cfloat<24,5> floor).
- Passes gcc + clang locally and in CI (CI_LITE builds elreal), including on a long-double-aliased target conceptually (the dyadic oracle makes the assertion platform-independent).

## References
- Epic: #923 (McCleeary LFPERA elreal). Blocks: #929 (Phase 5), and by extension #930-#933.
- Phases done: #925, #926, #927, #928; EFT exactness fix #942 (PR #1019).
- Oracle precedent: ereal epic #987, `include/sw/universal/verification/dyadic_exact.hpp`, `elastic/ereal/arithmetic/exact_value_oracle.cpp`.
- Rationale: `docs/assessments/ereal-retrospective.md`.

Ravenwater:
## Summary
The **Phase-5 precondition** (#1022): lock down a *validated* baseline for the elreal McCleeary substrate before building more arithmetic on it. Ports the ereal/#987 exact-oracle approach (`dyadic_exact.hpp`, an einteger-backed exact dyadic-rational reference sharing no code with the algorithms under test) to elreal, removing the Phase 1-4 reliance on `long double` -- which is not an exact oracle (it aliases `double` on MSVC/ARM/RISC-V, giving **zero** headroom over the type under test).

## What the oracle asserts
`exact(block) = from_double(double(b.v)) << b.exp`; `exact(ZBCL) = sum of block dyadics`.
- **block_two_sum / block_two_mult**: `exact(high)+exact(low) == exact(a){+,*}exact(b)` bit-for-bit, tolerating a nonzero error **only** when the residual block is a nonzero subnormal (the cfloat<24,5> #942 floor). A wrong high part or a wrong *normal* residual still fails.
- **threeAdd / add()**: exact dyadic equality on the wide-exponent RN hosts.
- Randomized + adversarial sweeps (cancellation, scale separation, classic `0.1+0.2`).

**Strict-exact PASSES on all five round-to-nearest hosts**: `float`, `double`, `half`, `cfloat<24,5>` (modulo #942), `cfloat<32,8>`.

## Finding (what the loose tolerances were masking)
**bfloat16 is not an exact-EFT host.** It rounds toward zero (`numeric_limits<bfloat16>::round_style == round_toward_zero`; its float->bf16 cast is `bits >> 16`, no rounding). Knuth/Dekker EFTs are exact only under round-to-nearest, so on bfloat16 they leak up to ~1 residual-ulp. This is **intentional bfloat16 behaviour, not an elreal bug** -- it makes elreal-over-bfloat16 inherently approximate (like the cfloat<24,5> floor, but caused by the rounding mode, not the exponent range). bfloat16 is **pinned** with a truncation bound (using `double` as its exact reference -- ~45 bits headroom over bfloat16's 8) so a regression is caught and the limitation is executable. A note is posted on epic #923.

## Changes
- `elastic/elreal/arithmetic/exact_value_oracle.cpp` (new) -- the keystone independent-oracle test.
- `elastic/elreal/arithmetic/addition.cpp` -- replaced the 4-ulp long-double band with the dyadic exact check; this also corrects a latent reference bug the band hid (for the `float` host the old reference summed the original `double` literals, not the represented float-cast operands).

## Test Results
| el_arith_exact_value_oracle | OK | PASS | OK | PASS |
| el_arith_addition | OK | PASS | OK | PASS |
| full elreal suite (16 tests) | OK | PASS | OK | PASS |

## Test plan
- [ ] Fast CI passes (gcc + clang CI_LITE builds elreal)

Resolves #1022
Relates to #923

## Summary by CodeRabbit

* **Tests**
  * Replaced approximate long-double checks with an exact dyadic-rational oracle for addition tests, improving precision and diagnostics.
  * Added a standalone exact-value verification tool to validate block-level and multi-block arithmetic across host types, with special handling for subnormal results and truncated-host behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/stillwater-sc/universal/pull/1023?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/elastic/elreal/arithmetic/addition.cpp
+++ b/elastic/elreal/arithmetic/addition.cpp
@@ -23,10 +23,16 @@
 namespace {
 
 // Exact value of a block as a dyadic rational (value(b) = v * 2^exp), shared
-// with the #1022 oracle. double(b.v) is exact for every elreal FpType.
+// with the #1022 oracle. double(b.v) is lossless only for an FpType with <= 53
+// significand bits; every elreal block host used here qualifies. A wider host
+// (e.g. cfloat<nbits,es,uint64_t,...> with nbits > 64) would need a wider exact
+// conversion, so we static_assert against it rather than silently mis-reference.
 template <typename FpType>
 sw::universal::dyadic exact_block(const sw::universal::block<FpType>& b) {
     using namespace sw::universal;
+    static_assert(std::numeric_limits<FpType>::digits <= 53,
+        "exact_block uses double(b.v) as the block's exact value, lossless only "
+        "for FpType with <= 53 significand bits.");
     if (b.is_zero_block()) return dyadic();
     dyadic d = dyadic::from_double(static_cast<double>(b.v));
     d.scale += b.exp;
--- a/elastic/elreal/arithmetic/addition.cpp
+++ b/elastic/elreal/arithmetic/addition.cpp
@@ -76,7 +76,10 @@ template <typename FpType>
 sw::universal::dyadic exact_value(const sw::universal::ZBCL<FpType>& z) {
     using namespace sw::universal;
     dyadic acc;
-    for (const auto& blk : z.take(16)) acc = acc + exact_block(blk);
+    // 32-block window, matching the #1022 oracle's ZBCL_EXACT_WINDOW so both
+    // files agree on the exact value of the same stream; finite test sums settle
+    // well within it.
+    for (const auto& blk : z.take(32)) acc = acc + exact_block(blk);
     return acc;
 }
 
--- a/elastic/elreal/arithmetic/addition.cpp
+++ b/elastic/elreal/arithmetic/addition.cpp
@@ -17,31 +17,48 @@
 #include <iostream>
 
 #include <universal/number/elreal/elreal.hpp>
+#include <universal/verification/dyadic_exact.hpp>
 #include <universal/verification/test_suite.hpp>
 
 namespace {
 
+// Exact value of a block as a dyadic rational (value(b) = v * 2^exp), shared
+// with the #1022 oracle. double(b.v) is exact for every elreal FpType.
 template <typename FpType>
-long double block_value(const sw::universal::block<FpType>& b) {
-    if (b.is_zero_block()) return 0.0L;
-    return static_cast<long double>(b.v) * std::ldexp(1.0L, b.exp);
+sw::universal::dyadic exact_block(const sw::universal::block<FpType>& b) {
+    using namespace sw::universal;
+    if (b.is_zero_block()) return dyadic();
+    dyadic d = dyadic::from_double(static_cast<double>(b.v));
+    d.scale += b.exp;
+    return d;
 }
 
+template <typename FpType>
+sw::universal::dyadic exact_value(const sw::universal::ZBCL<FpType>& z) {
+    using namespace sw::universal;
+    dyadic acc;
+    for (const auto& blk : z.take(16)) acc = acc + exact_block(blk);
+    return acc;
+}
+
+// Value preservation, validated against an INDEPENDENT exact dyadic oracle
+// (not long double -- see #1022). double and float are round-to-nearest, so
+// add() must reproduce the exact sum of the REPRESENTED operands bit-for-bit:
+//   exact(add(za, zb)) == exact(za) + exact(zb).
+// The reference uses the represented values exact(za)/exact(zb), not the
+// original double literals, so it is correct for the float host too (the old
+// 4-ulp band masked that distinction).
 template <typename FpType>
 int verify_value(double a, double b, const std::string& tag) {
     using namespace sw::universal;
     int nrFailures = 0;
-    auto z = add(from_native<FpType>(a), from_native<FpType>(b));
-    auto blocks = z.take(8);
-    long double got = 0.0L;
-    for (const auto& blk : blocks) got += block_value(blk);
-    long double ref = static_cast<long double>(a) + static_cast<long double>(b);
-    constexpr int p = std::numeric_limits<FpType>::digits;
-    long double scale = std::fmax(std::fabs(ref), 1.0L);
-    long double tol = scale * std::ldexp(1.0L, -p) * 4; // a few FpType-ulps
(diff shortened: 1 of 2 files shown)


elastic/elreal/arithmetic/addition.cpp
```diff
@@ -17,31 +17,48 @@
 #include <iostream>
 
 #include <universal/number/elreal/elreal.hpp>
+#include <universal/verification/dyadic_exact.hpp>
 #include <universal/verification/test_suite.hpp>
 
 namespace {
 
+// Exact value of a block as a dyadic rational (value(b) = v * 2^exp), shared
+// with the #1022 oracle. double(b.v) is exact for every elreal FpType.
```
Ravenwater: that is not true for cfloat<nbits, es, std::uint64_t, true, false, false> for anything with nbits > 64.

Ravenwater: Good catch -- corrected in c707682b. The comment now states the real condition (double(b.v) is lossless only for <= 53 significand bits, which every block host used here meets), and both exact_block() definitions now static_assert(std::numeric_limits<FpType>::digits <= 53, ...) so a wider host such as cfloat<nbits,es,uint64_t,...> with nbits > 64 fails to compile rather than silently producing a wrong dyadic reference.

Issue #931, opened 2026-05-22
elreal Phase 7: transcendental functions and numbers


Ravenwater:
**Parent epic: #923**
**Depends on: Phase 6 (multiplication / division)**

## Goal

Provide the math suite: \`exp\`, \`log\`, \`sin\`, \`cos\`, \`tan\`, \`sqrt\`, \`atan\`, \`asin\`, \`acos\`, \`sinh\`, \`cosh\`, \`tanh\`, \`pow\`, \`hypot\`. Plus the named constants: \`pi\`, \`e\`, \`ln(2)\`, \`log2(10)\`, \`euler_gamma\`, \`phi\`, \`sqrt(2)\`, \`sqrt(3)\`, \`sqrt(5)\`.

## Algorithm choices (per dissertation 4.2.7)

- Square root: Newton-Raphson on \`ZBCL\`
- exp: Taylor series, with argument reduction \`exp(x) = exp(x/2^k)^(2^k)\`
- log: Taylor series of \`ln(1+u)\` with argument reduction \`log(2^k * m) = k*ln(2) + log(m)\`
- sin / cos: Taylor series with Payne-Hanek-style range reduction modulo 2*pi (using \`pi\` itself as a ZBCL)
- atan: Machin formula or its extensions for fast pi convergence
- pi: BBP digit extraction, Machin, or AGM (the dissertation discusses options)
- e: Taylor series of \`exp(1)\` -- direct
- ln(2): Mercator series or related

## Where exp_offset earns its keep

Transcendentals at extreme arguments (\`exp(1000)\`, \`log(1e-300)\`) have combined exponents far beyond any single host FP type's range. The \`exp_offset\` field on \`block<FpType>\` lets a single block carry that exponent. Phase 7 will be the first phase where this matters in practice.

## Acceptance criteria

- [ ] All listed functions implemented, each its own \`.hpp\` under \`include/sw/universal/number/elreal/math/\`
- [ ] All listed constants are lazy \`ZBCL<FpType>\` factories (not eagerly materialised double constants)
- [ ] Tests check first 100 decimal digits of each constant against the 320-digit reference strings in \`include/sw/math/constants/reference_constants.hpp\` (PR #915)
- [ ] Function tests: roundtrip identities -- \`log(exp(x)) == x\`, \`exp(log(x)) == x\` for x > 0, \`sin(asin(x)) == x\` on \`[-1,1]\`, etc., stream-wise to depth N
[...]

PR #1039, opened 2026-06-02
feat(elreal): named constants as lazy ZBCL factories (Phase 7.1, #931)

Ravenwater:
## Summary
Sub-phase **7.1** of Phase 7 (Epic #923): the named constants as **generated** lazy ZBCL factories (not baked-in double literals), building on Phase 5/6 and 7.2 (sqrt).

| Constant | Method |
|----------|--------|
| `pi` | Machin: `16*atan(1/5) - 4*atan(1/239)` |
| `e` | exp(1) Taylor series `sum 1/n!` |
| `ln2` | `2*artanh(1/3)`; `ln10` = `3*ln2 + 2*artanh(1/9)` |
| `log2(10)` | `ln10 / ln2` |
| `sqrt2/3/5` | Newton-Raphson sqrt (7.2); `phi` = `(1+sqrt5)/2` |
| `euler_gamma` | host-precision seed (no efficient elementary series — the one exception; high-precision expansion is a follow-up) |

`detail::odd_power_series` is the shared atan/artanh odd-power series, clamped a margin above the denormal floor.

## Host scope (deliberate)
- **Series constants** (pi/e/ln2/ln10/log2(10)) — tested on `{double, float}`. **bfloat16's 8-bit exponent range cannot hold** the deep expansions these series generate (it can't even represent the partial sums), so they're out of scope for bfloat16.
- **Radicals + euler_gamma** — `{double, float, bfloat16}` (they stay in range).
- **Beyond-double precision:** on a double host, `pi` agrees with a 3-limb high-precision reference to **> 100 bits** (concrete proof of precision past the host double). Full 100-digit reference-string verification is a follow-up (needs a decimal→dyadic parser).

## Notable fixes (all surfaced only with assertions on — i.e. CI's Debug sanitizer/coverage jobs)
- bfloat16 series underflow → scoped series constants to double/float.
- high-precision check read a leading **zero block**'s exponent → switched to residual magnitude.
- `e_zbcl<float>` fed a **subnormal** `1/n!` to `from_native` (asserts) → bounded the factorial loop at `min_normal`.

## Test results (local)
| Suite | gcc | clang | asserts | RISC-V |
|-------|-----|-------|---------|--------|
| Full elreal ctest (28 tests, +constants) | 28/28 | 28/28 | pass | cross-compiles clean |

ASCII Guard clean.

## Test plan
- [ ] Fast CI (gcc + clang CI_LITE builds elreal)
- [ ] Promote to ready

Relates to #931 (sub-phase 7.1).

## Summary by CodeRabbit

* **New Features**
  * Added lazily-generated, high-precision mathematical constants (e, ln2, ln10, log2(10), pi, sqrt2, sqrt3, sqrt5, phi, euler gamma) with support for multiple floating-point types.

* **Tests**
  * Added a comprehensive validation suite that verifies constant accuracy and numeric properties across precisions and reports failures.


Ravenwater:
## Correction: the bfloat16 rationale in the summary was wrong

Good catch on the review of this thread. The PR summary said the series constants are excluded from bfloat16 because of its "8-bit exponent range" — that's **incorrect**: bfloat16 has the **same 8-bit exponent range as float32** (they differ only in the significand, 7 vs 23 bits). Fixed the comment in `constants.cpp` (commit `a6eeabda`).

**The actual reason** (measured):

| host | k (significand) | `1/5` via div | square in `mul` | result |
|------|------|------|------|------|
| float | 24 | ~3 blocks, down to ~2^-55 | ~2^-110 (in range) | **works** |
| bfloat16 | 7 | ~15 blocks, down to ~2^-115 (near floor) | ~2^-230 (**below representable**) | **underflows/aborts at depth 16** |

So it's bfloat16's **narrow 7-bit significand**, not its exponent range: with k=7, a division-derived operand like `1/5` spans ~15 blocks reaching near the exponent floor, and squaring it in `mul()` lands far below the smallest representable value.

At a *small* depth bfloat16 series do run without aborting, but its ~7-bit arithmetic caps accuracy at ~2-3 digits and **degrades** with more depth (e.g. pi<bf>: 3.141 at depth 3, 3.175 at depth 4+), so high-precision series constants aren't meaningful on bfloat16 regardless. The scoping (series → {double,float}; radicals/euler_gamma → all three) stands; only the explanation was wrong.

PR #1134, opened 2026-07-17
fix(bfloat16): round float conversion to nearest-even (RNE) instead of truncating

### LINKED ISSUE #1133: Rounding mode for BFloat16
It appears the rounding mode is "round to 0" (truncate), but Google TPUs are using RNE, and Intel followed suit.

Here is a small test:

0.2691408770292272 --> 0.267578125  while same value with RNE = 0.26953125.

Ravenwater:
## Summary
- `bfloat16`'s float/double conversion was **truncating** (round-toward-zero): `convert_ieee754()` did a bare `bits >> 16`, dropping the low 16 bits of the IEEE-754 float with no rounding.
- Google TPUs and Intel round float->bfloat16 to **nearest, ties-to-even (RNE)**. This PR implements RNE.

## Root cause
`include/sw/universal/number/bfloat16/bfloat16_impl.hpp` kept the upper 16 bits of the source float verbatim. For the issue's value `0.2691408770292272` (float bits `0x3E89CCD5`), the dropped bits `0xCCD5` are more than half a ulp, so RNE must round **up**:
- truncation -> `0.267578125`
- RNE (correct) -> `0.26953125`

## Changes
- `include/sw/universal/number/bfloat16/bfloat16_impl.hpp` -- RNE via the standard magic rounding bias `0x7FFF + lsb_of_retained_field` before the shift, plus a NaN guard so a float NaN whose payload lives only in the low 16 bits does not collapse to +/-inf. Carry into the exponent and round-to-inf at the range limit both fall out correctly.
- `include/sw/universal/number/bfloat16/numeric_limits.hpp` -- `round_style` now `round_to_nearest` (was `round_toward_zero`).
- `static/float/bfloat16/conversion/round_to_nearest_even.cpp` -- new regression test: the issue case, more-than-half/less-than-half rounding, exact-half tie-to-even (lsb=0 stays, lsb=1 rounds up), and inf/nan/zero preservation.

## Test Results (built with `-I include/sw`, `-std=c++20`)
| bfloat16_round_to_nearest_even (new) | OK | PASS | OK | PASS |
| bfloat16_string_parse | OK | PASS | OK | PASS |
| bfloat16_arithmetic | OK | PASS | OK | PASS |
| bfloat16_api | OK | PASS | OK | PASS |

gcc 13.3.0, clang 18.1.3.

## Test plan
- [ ] Promote to ready when satisfied: `gh pr ready`

Resolves #1133

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved bfloat16 down-conversion accuracy, using round-to-nearest, ties-to-even behavior.
  - Fixed halfway/tie rounding so results match expected bfloat16 rounding semantics.
  - Preserved NaNs (including payloads) and continued correct handling of zeros and infinities.
- **Compatibility**
  - Updated bfloat16 rounding metadata to report round-to-nearest behavior.
- **Tests**
  - Added regression coverage to validate round-to-nearest-even conversions and special-value behavior.

### DIFF
--- a/include/sw/universal/number/bfloat16/bfloat16_impl.hpp
+++ b/include/sw/universal/number/bfloat16/bfloat16_impl.hpp
@@ -61,9 +61,26 @@ class bfloat16 {
 		typename = typename std::enable_if< std::is_floating_point<Real>::value, Real >::type>
 	BIT_CAST_CONSTEXPR bfloat16& convert_ieee754(Real rhs) noexcept {
 		// Down-cast wider floats to single precision first; bfloat16 is
-		// defined relative to the 32-bit IEEE-754 layout.
+		// defined relative to the 32-bit IEEE-754 layout. bfloat16 keeps the
+		// top 16 bits (sign + 8 exponent + 7 fraction) and must round the
+		// discarded low 16 bits to nearest, ties-to-even (RNE), matching the
+		// hardware behavior of Google TPUs and Intel. See issue #1133.
 		float f = static_cast<float>(rhs);
 		uint32_t bits = sw::bit_cast<uint32_t>(f);
+		// A float NaN whose payload lives only in the low 16 bits would, after
+		// truncation or rounding, collapse to +/-inf. Preserve NaN by forcing a
+		// quiet-NaN fraction bit while keeping the sign and exponent.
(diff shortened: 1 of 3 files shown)


jamesquinlan (approved):
LGTM. Thanks for taking the time to put this together.

PR #1138, opened 2026-07-18
fix(efloat): implement abs()/fabs() (were a no-op stub)

Ravenwater:
## Summary
The free function `abs(efloat<nlimbs>)` was a stub that returned its argument unchanged (`return a;`), so `abs(-x) == -x`. Every caller expecting a magnitude got a signed value. This also adds the standard `<cmath>` spelling `fabs()`.

## Root cause
```cpp
template<unsigned nlimbs>
inline efloat<nlimbs> abs(const efloat<nlimbs>& a) {
	return a; // (a < 0 ? -a : a);   <-- never implemented
}
```

## Fix
Clear the sign on a copy -- correct for every state: negative normal -> positive, `-inf -> +inf`, `-0 -> +0`, NaN stays NaN (sign is irrelevant to `isnan()`). `fabs` is a thin alias.

## Side benefit (log1p)
`log1p`'s small-argument guard `abs(x) < 0.375` (`logarithm.hpp`) was effectively the **signed** test `x < 0.375` under the stub, routing moderately negative arguments into the slow Taylor path. With a correct `abs` it now gates by **magnitude** as intended. Verified `log1p` matches `std::log1p` to machine precision for `x in [-0.9, 0.9]`.

## Changes
- `include/sw/universal/number/efloat/efloat_impl.hpp` -- real `abs()`, new `fabs()`.
- `elastic/efloat/math/absolute.cpp` (new) -- regression test (target `efloat_absolute`) covering normal/zero/inf/NaN and explicitly asserting `abs(-x) != x`.

## Test Results
| efloat_absolute (new) | OK | PASS | OK | PASS |
| efloat_math / fractional / minmax / classify / pow / exponent (regression) | OK | PASS | -- | -- |

gcc 13.3.0, clang 18.1.3.

## Test plan

### DIFF
--- a/elastic/efloat/math/absolute.cpp
+++ b/elastic/efloat/math/absolute.cpp
@@ -1,0 +1,124 @@
+// absolute.cpp: regression tests for efloat abs() / fabs().
+//
+// Guards against the prior stub that returned its argument unchanged (so
+// abs(-x) == -x). Verifies sign clearing across normal, zero, inf, and NaN.
+//
+// Copyright (C) 2017 Stillwater Supercomputing, Inc.
+// SPDX-License-Identifier: MIT
+//
+// This file is part of the universal numbers project, which is released under an MIT Open Source license.
+#include <universal/utility/directives.hpp>
+#include <cmath>
+#include <limits>
+#include <universal/number/efloat/efloat.hpp>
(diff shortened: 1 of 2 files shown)


Ravenwater:
@coderabbitai review

Issue #1141, opened 2026-07-18
efloat: parse() silently returns ~0 when target precision exceeds the 2040-bit d2b cap


Ravenwater:
## Summary
`parse()` into an `efloat` whose `get_precision()` exceeds **2040 bits** returns `valid == true` but yields a **wrong value (~0)** rather than either capping gracefully or reporting failure.

## Reproduction
```cpp
efloat<128> pi; pi.set_precision(3400);   // > 2040
bool ok = parse("3.14159265358979323846...", pi);
// ok == true, but double(pi) == 0.0  (garbage)
```

## Root cause
`efloat_impl.hpp::parse()` sizes the d2b target as `target_bits = min(get_precision(), default_big_bits - 8)` = `min(want, 2040)`, then packs the mantissa MSB-first into `get_precision()`-worth of limbs. When `get_precision() > 2040`, the limb packing and exponent bookkeeping are inconsistent with the 2040-bit `convert()` result, producing ~0. `decimal_to_binary::convert` is itself template-parameterized on `BigBits` (default 2048), so the ceiling is a *default*, not a hard limit -- `convert<8192>(...)` yields correct values well past 1000 digits (verified while scoping #1139).

## Proposed fix
Either (a) template `parse()` on `BigBits` (default `default_big_bits`) so callers can request a larger exact budget and pack `min(get_precision(), BigBits-8)` bits consistently, or (b) if keeping the 2040 cap, clamp `get_precision()` to the cap for the packing too (so the result is a correctly-rounded 2040-bit value, not garbage) and/or return `false`. Option (a) is the enabler for #1139's 1000-digit constants.

## Impact
Silent wrong results for any high-precision decimal literal; blocks 1000-digit constants (#1139). Relates to Epic #1101.

Issue #1140, opened 2026-07-18
efloat: sqrt() accuracy frozen at ~97 digits regardless of get_precision()


Ravenwater:
## Summary
`efloat<nlimbs>::sqrt` produces a result accurate to only **~323 bits (~97 decimal digits)** no matter how high the operand's precision is set. Requested precision has no effect on accuracy.

## Reproduction
```cpp
for (unsigned bits : {512u, 1024u, 2048u, 3400u}) {
    efloat<128> two(2.0); two.set_precision(bits);
    efloat<128> r = sqrt(two);
    efloat<128> e = r*r - two; e.setsign(false);
    // |r*r - 2| scale is ALWAYS ~ -323, for every `bits`
}
```
Output: `|r*r-2|` binary scale is `-323` at 512, 1024, 2048, and 3400 bits alike.

## Root cause
`math/sqrt.hpp` runs a **fixed** number of Newton iterations (`for (int i = 0; i < 7; ++i)`). Seven iterations from the ~52-bit double seed cannot reach beyond a few hundred bits in practice, and the count does not scale with `get_precision()`. (By contrast, efloat's `+ - * /` are correct to 3400+ bits -- verified `3*(1/3)-1` at scale -3425 -- so the limitation is specific to the sqrt/Newton kernels.)

## Proposed fix
Make the iteration count precision-adaptive: iterate `ceil(log2(target_bits / 52)) + guard` times (Newton doubling), or loop until the correction term drops below the working precision (the `term.scale() < result.scale() - get_precision()` idiom the Taylor kernels use). Same likely applies to `cbrt` and any other fixed-iteration kernel.

## Impact
Blocks high-precision `sqrt`-based work, including native computation of constants like phi = (1+sqrt(5))/2 and pi via Gauss-Legendre AGM (surfaced while scoping #1139). Relates to Epic #1101.

PR #1137, opened 2026-07-18
feat(efloat): implement math/trigonometry.hpp (sin, cos, tan, asin, acos, atan, atan2)

Ravenwater:
## Summary
Implements `efloat/math/trigonometry.hpp` for the arbitrary-precision `efloat` template, the last missing piece of the efloat mathematical library (parent #1092). These are **native arbitrary-precision** implementations (argument reduction + series), not double-delegating shims.

## Changes
- `include/sw/universal/number/efloat/math/trigonometry.hpp` (new) -- `sin`, `cos`, `tan`, `asin`, `acos`, `atan`, `atan2` in `sw::universal`, plus internal `efloat_pi`/`efloat_pi_2`/`efloat_pi_4` helpers.
- `include/sw/universal/number/efloat/mathlib.hpp` -- include the new header after `truncate.hpp` (trig needs `round()`).
- `elastic/efloat/math/trigonometry.cpp` (new) -- regression test (auto-globbed as target `efloat_trigonometry`).

## Algorithm
- **sin/cos**: reduce to `[-pi, pi]` via `r = x - round(x/2pi)*2pi`, then Taylor series with the precision-aware `term.scale() < sum.scale() - get_precision()` terminator (same idiom as `expm1`/`log1p`).
- **tan** = sin/cos, signed-infinity at the pole (`cos == 0`).
- **atan**: `|x|==1 -> pi/4`; `|x|>1 -> pi/2 - atan(1/|x|)`; `0.5<|x|<=1 ->` addition formula `atan(1/2) + atan((|x|-1/2)/(1+|x|/2))`; else Taylor -- keeps the reduced argument small so the series converges fast (no slow Leibniz tail near 1).
- **asin**: domain `|x|>1 -> NaN`; `|x|>0.8 -> pi/2 - asin(sqrt(1-x^2))`; else Taylor.
- **acos** = `pi/2 - asin`.
- **atan2**: full-quadrant, using the efloat pi constants (**not** bare `double` literals -- avoids the double-contamination bug present in `ereal`'s `atan2`).

## Notes / follow-ups
- High-precision pi is seeded via `parse()` of a ~62-digit literal (the same idiom `exp`/`log` use for `ln2`); `pi/2`, `pi/4` derived by exact multiplication. Accuracy is therefore capped at ~62 digits like the existing `ln2`/`ln10` constants -- fine for the standard efloat configs; a longer literal or a native pi computation can lift that in a later PR.
[...]

PR #1142, opened 2026-07-18
feat(efloat): 1000-digit pi, e, phi constants as highest-precision basis

### LINKED ISSUE #1139: feat(efloat): 1000-digit constants (pi, phi, e) as the library's highest-precision constant basis
## Goal
Provide **1000-decimal-digit** (~3322-bit) constants for **pi**, **phi** (golden ratio), and **e** in the `efloat` type, so `efloat` becomes the most precise constant basis in the library -- roughly **3x more accurate than any current multi-component representation** (`ereal`/`elreal` top out at ~315 digits; see below).

Relates to Epic #1101 (Complete efloat as an Oracle-Grade Arbitrary-Precision Type).

## Motivation
`efloat` is a single-component, arbitrary-precision binary float, so its precision scales directly with `nlimbs` (32 bits/limb). At **~104 limbs** it holds ~3322 bits = ~1000 decimal digits. But today there is no constant anywhere in `efloat` that carries that many digits:

- Existing efloat constants are **~60-digit decimal literals** parsed at runtime: `ln2`/`ln10` (in `math/exponent.hpp`/`math/logarithm.hpp`) and the `pi` introduced by the trigonometry work (PR #1137, `math/trigonometry.hpp`). These cap accuracy at ~60 digits regardless of `nlimbs`.
- The multi-component types are better but still bounded: `ereal`/`elreal` store pi/e/etc. as **19-limb non-overlapping double expansions, ~315 decimal digits**, generated offline with MPFR and verified (`ereal/math/constants/ereal_constants.hpp:63,75`).

A 1000-digit basis (~3.2x the 315-digit ceiling) would let `efloat` back the highest-fidelity reproducible computations in the library (transcendental identities, oracle comparisons, argument reduction for very-high-precision trig/exp/log).

## Known blocker: the parse() precision ceiling
`efloat`'s decimal `parse()` path is **capped at 2040 bits** (~614 digits):
```
// efloat_impl.hpp:1289-1291
// The d2b target_mantissa_bits is ... capped at 2040 bits (just under the d2b BigBits budget of 2048).
```
`decimal_to_binary::default_big_bits` is 2048 and the cap is `default_big_bits - 8`. So **parse() alone cannot reach 1000 digits today** -- the budget must be raised (to >= ~3400 bits) or an alternate construction used. This is the central engineering task of this issue, not just adding string literals.

## Proposed scope
1. **Raise the `decimal_to_binary` BigBits budget** (`default_big_bits`) enough to represent >= 3400 bits, or provide a chunked/native construction path that bypasses the 2040-bit cap. Confirm no regression to the existing d2b consumers (dfloat, ereal parse, etc.).
2. **Add `efloat/math/constants/efloat_constants.hpp`** with `efloat_pi<nlimbs>()`, `efloat_e<nlimbs>()`, `efloat_phi<nlimbs>()` (matching the `ereal_constants.hpp` free-function idiom), seeded from **1000-digit literals** (pi: OEIS A000796; e: A001113; phi: A001622).
3. **Refactor PR #1137's inline ~62-digit pi** (`math/trigonometry.hpp`) to consume `efloat_pi<nlimbs>()`, so trig inherits the full-precision constant.
4. **Validation** against an MPFR-generated 1000-digit reference (a generator under `tools/generators/`), gated behind `REGRESSION_LEVEL_4` (the pattern the elreal 320-digit constant check uses), so default builds stay fast.

## Acceptance criteria
- [ ] `efloat<104>` (or the config that yields >= 3322 bits) reproduces pi, e, phi to **>= 1000 correct decimal digits**, verified against MPFR.
- [ ] parse() (or the chosen construction) can represent >= 3400-bit values; existing d2b consumers unaffected.
- [ ] `efloat_pi`/`efloat_e`/`efloat_phi` available via a constants header and used by the trig library.
- [ ] LEVEL_4 regression test validating all three constants to 1000 digits; default-level smoke test to ~60+ digits.

## Notes
- 1000 digits ~= 3322 bits ~= 104 x uint32 limbs.
- phi = (1 + sqrt(5))/2 could alternatively be computed natively from a 1000-digit sqrt(5); e via its rapidly-converging series; pi via Chudnovsky/AGM -- native computation is an option that sidesteps the literal-length/parse-ceiling issue entirely and is worth weighing against the literal approach.
- Depends on / follows PR #1137 (efloat trigonometry, which introduced the first efloat pi) and the abs()/fabs() fix (PR #1138).

Ravenwater:
## Summary
Adds `efloat/math/constants/efloat_constants.hpp` with `efloat_pi<nlimbs>()`, `efloat_e<nlimbs>()`, `efloat_phi<nlimbs>()`, seeded from **~1000-digit** decimal literals. A ~3300-bit `efloat<128>` reproduces each to **~1000 decimal digits** -- roughly **3x** the ~315-digit ceiling of the multi-component `ereal`/`elreal` constants (issue #1139).

## Approach (Path A: literal + wide-convert)
Construction **bypasses `efloat::parse()`** (capped at the 2040-bit d2b default, tracked in #1141) and calls `decimal_to_binary::convert<16384>` directly, then packs the mantissa MSB-first into the efloat at `min(nlimbs*32, ~3300)` bits.

The large `BigBits=16384` is **required**: the d2b reduction shifts the digit-integer left by ~`(target + 3*ndigits)` bits, so for a 1000-digit literal at a ~3300-bit target the intermediate needs ~9700 bits. This does **not** change the shared d2b default budget -- `dfloat`/`dd`/`qd`/`cfloat`/`posit` are unaffected. Values are cached per instantiation via a `static` local.

## Changes
- `include/sw/universal/number/efloat/math/constants/efloat_constants.hpp` (new) -- constants + wide-parse helper.
- `include/sw/universal/number/efloat/math/trigonometry.hpp` -- consumes the new `efloat_pi` (drops its inline ~62-digit literal), addressing the arbitrary-precision-fidelity concern from the #1137 review.
- `include/sw/universal/number/efloat/mathlib.hpp` -- include the constants header before trigonometry.
- `elastic/efloat/math/constants.cpp` (new) -- regression test (target `efloat_constants`).
- `tools/generators/efloat_constants_gen.py` (new) -- mpmath generator for the literals.

## Validation (independent oracles, not the literal as its own reference)
- **phi** -- identity `phi^2 - phi - 1 == 0`.
- **e** -- vs the native series `sum 1/k!` (uses only `+,*,/`).
- **pi** -- vs Machin's `16*atan(1/5) - 4*atan(1/239)` (atan of a small argument is a pure Taylor series and does not use pi).

Level 1 validates ~308 digits (`efloat<32>`, fast); **LEVEL_4** validates the full **~1000 digits** (`efloat<128>`). All pass on gcc 13.3.0 and clang 18.1.3; trig regression still passes.

## Test Results
| efloat_constants L1 (~308 digits) | OK | PASS | OK | PASS |
| efloat_constants L4 (~1000 digits) | OK | PASS | OK | PASS |
| efloat_trigonometry (regression) | OK | PASS | OK | PASS |

## Related / follow-ups
- Surfaced two efloat bugs, filed separately: **#1140** (sqrt accuracy frozen at ~97 digits) and **#1141** (parse() >2040 bits returns garbage). This PR sidesteps both via the wide-convert path; fixing #1141 would later let `efloat_pi` be built through `parse()` directly.
- Native computation (Gauss-Legendre pi, etc.) was considered but is gated on #1140; the literal path here is exact and independent of it.

## Test plan

Resolves #1139


Ravenwater:
@coderabbitai review

PR #1144, opened 2026-07-18
fix(efloat): parse() high precision without silent garbage (+ BigBits budget)

### LINKED ISSUE #1141: efloat: parse() silently returns ~0 when target precision exceeds the 2040-bit d2b cap

Ravenwater:
## Summary
`efloat::parse()` silently returned **~0** once the requested precision approached the d2b working budget (issue #1141). Even a short literal like `pi` produced `0.0` at `set_precision(2040+)`.

## Root cause
`parse()` capped `target_bits` at `default_big_bits - 8 = 2040` and called `convert<2048>`. But `convert`'s reduction **left-shifts the digit-integer by ~`(target + 3*neg_E)` bits** (`neg_E` = fractional places + any negative scientific exponent). For a target anywhere near 2048 that shift **overflows** the `BigBits` integer -> garbage. So the 2040 cap was itself unsafe, independent of literal length.

## Fix
- **Template `parse` on `BigBits`** (default `decimal_to_binary::default_big_bits`). `parse(s,x)` is unchanged (backward compatible); callers can request a larger exact budget -- `parse<16384>(s,x)` -- to represent very high precision (e.g. 1000-digit constants). The round-up bigint now honors `BigBits` (was hardcoded to the default). The default lives on the forward declaration in `efloat_fwd.hpp` (what `assign()` sees), which now includes `decimal_to_binary.hpp`.
- **Overflow-safe `target_bits`**: from the scanned literal, compute the effective decimal exponent and significant-digit count, and cap `target` so the convert shift stays within `BigBits`. A request beyond the safe ceiling is now **correctly rounded to the ceiling** instead of returning garbage.

## Changes
- `include/sw/universal/number/efloat/efloat_impl.hpp` -- overflow-safe target sizing, `convert<BigBits>`, `BigBits`-correct round-up.
- `include/sw/universal/number/efloat/efloat_fwd.hpp` -- `parse` fwd decl carries the `BigBits` default; include d2b for `default_big_bits`.
- `elastic/efloat/conversion/parse_precision.cpp` (new) -- regression test (target `efloat_parse_precision`).

## Verification
- Default `parse` of short & long pi at precision 256..3400 now returns the **correct value** (was `~0` at >= 2040).
- `parse<16384>` of a 1000-digit pi matches `efloat_pi<128>()` to **~1000 digits**; `parse<16384>("0.1")*10 == 1` to **~3300 bits** (independent identity).
- All **22 existing efloat regression tests pass** on gcc 13.3.0 and clang 18.1.3.

## Notes
- This makes `parse<16384>` a general high-precision path -- the #1139 constants' bespoke `make_constant` helper could later be simplified to use it.

## Test plan

Resolves #1141

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved high-precision decimal parsing for `efloat` values.
  - Added configurable precision control for parsing very large or highly precise numbers.
  - Prevented precision loss and incorrect rounding during conversion.
  - Preserved correct handling of special values such as `nan`, `inf`, and zero.
  - Improved rejection of malformed numeric input.

- **Tests**
  - Added regression coverage for precision, rounding, long decimal literals, special values, and invalid input.

### DIFF
--- a/elastic/efloat/conversion/parse_precision.cpp
+++ b/elastic/efloat/conversion/parse_precision.cpp
@@ -1,0 +1,126 @@
+// parse_precision.cpp: regression tests for high-precision efloat parse().
+//
+// Guards issue #1141: parse() silently returned ~0 once the requested precision
+// approached the d2b working budget (the convert reduction's internal shift
+// overflowed BigBits). parse() now sizes its target overflow-safely, so a
+// high-precision request yields a correctly-rounded value; and parse<BigBits>
+// with a larger budget reaches arbitrary precision.
+//
+// Copyright (C) 2017 Stillwater Supercomputing, Inc.
+// SPDX-License-Identifier: MIT
+//
+// This file is part of the universal numbers project, which is released under an MIT Open Source license.
+#include <universal/utility/directives.hpp>
+#include <cmath>
+#include <string>
+#include <universal/number/efloat/efloat.hpp>
+#include <universal/verification/test_suite.hpp>
+
+namespace {
+
+	int VerifyEfloatParsePrecision(bool reportTestCases) {
+		using namespace sw::universal;
+		int failures = 0;
+
+		const std::string pi_short = "3.141592653589793";
+		const std::string pi_long  =
+			"3.14159265358979323846264338327950288419716939937510582097494459230781"
+			"640628620899862803482534211706798214808651328230664709384460955058223";
+		const double PI = std::acos(-1.0);
+
+		// ---------------------------------------------------------------------
+		// 1. #1141 guard: default parse() must return the correct value (not ~0)
+		//    at every requested precision, for short and long literals.
+		// ---------------------------------------------------------------------
+		if (reportTestCases) std::cout << "  Verifying default parse() is correct at high precision...\n";
+		{
+			for (unsigned prec : { 256u, 512u, 2040u, 2048u, 3400u }) {
+				efloat<128> a; a.set_precision(prec);
+				efloat<128> b; b.set_precision(prec);
+				bool oka = parse(pi_short, a);
+				bool okb = parse(pi_long, b);
+				if (!oka || std::abs(double(a) - PI) > 1e-10) { if (reportTestCases) std::cout << "    FAIL: parse(pi_short)@" << prec << " = " << double(a) << "\n"; ++failures; }
+				if (!okb || std::abs(double(b) - PI) > 1e-10) { if (reportTestCases) std::cout << "    FAIL: parse(pi_long)@"  << prec << " = " << double(b) << "\n"; ++failures; }
+			}
+		}
+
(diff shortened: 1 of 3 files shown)


Ravenwater:
### CodeRabbit review disposition (commit 58b91a81)

**Addressed both findings:**
- **Major -- positive scientific exponents not budgeted:** correct and important. `convert<BigBits>` grows the digit-integer by ~`E*log2(5)` bits (the `5^E` factor) for `E > 0`, which the old ceiling ignored, so `1e1000` could still overflow -> garbage. Now budgeting both directions: reject (return `false`) when the digit-integer plus its `5^|E|` growth can't fit in `BigBits`, and size `target` overflow-safely (target-relative for `E<0`, magnitude-only for `E>=0`). `1e1000` now fails cleanly at the default budget and parses correctly under `parse<16384>` (~`2^3321.9`). Added `1e1000` regression coverage.
- **Minor -- assert accuracy not capacity:** the `parse<16384>(long-pi)` test now checks actual accuracy against an independent reference (`efloat_pi`, ~430-bit agreement for the 130-digit literal) instead of `get_precision()` capacity.

Normal/scientific literals (`1.25e3`, `6.022e23`, `1e-100`, ...) are unaffected; all 22 efloat regression tests pass on gcc + clang.

Issue #1209, opened 2026-07-23
Phase 4: deprecate + fully remove the linear-algebra code from Universal


Ravenwater:
Sub-issue of #1204. **Target repo: universal.** Depends on Phases 1-3 (all consumers re-homed).

End state (decision #1): Universal is a **number-systems-only** library.

## Tasks
- [x] One-release **deprecation window**: add `#warning`/pragma-deprecation to `blas/blas.hpp` and `numeric/containers.hpp` pointing to MTL5/mp-*, and a CI note. Cut a release. _(deprecation notices + migration doc landed in PR #1212; **release cut still pending** before the removal below.)_
- [ ] Then **remove**: `include/sw/blas/**`, `include/sw/numeric/containers/**`, `linalg/**`, `benchmark/*/blas`, the BLAS-consuming `applications/*` and `mixedprecision/**`, and the `blas/matrices/` test-matrix DB (move any still-needed matrices to MTL5 first).
- [ ] Remove CMake options + subdir adds: `UNIVERSAL_BUILD_LINEAR_ALGEBRA_{BLAS,VMATH,DATA}`, `UNIVERSAL_BUILD_MIXEDPRECISION_SDK` (+ sub-options), and the `linalg`/`benchmark/*/blas`/`applications`/`mixedprecision` `add_subdirectory` lines.
- [ ] Update docs (README, docs/) to point linear-algebra users at MTL5 + mp-*.
- [ ] Confirm the number-system regression suites and remaining tools build with zero references to `blas/` or `numeric/containers/`.

## Acceptance
- [ ] Universal builds and all number-system tests pass with the linear-algebra code fully removed.
- [ ] No dangling includes/options; docs updated.

Closes out #1204.

Ravenwater:
Phase 4 started -- the **deprecation-window** step (PR #1212, draft):
- `blas/blas.hpp` + `numeric/containers.hpp` emit a `#pragma message` on include pointing to MTL5 + mp-*; informational (does not fail -Werror), silence with `-DUNIVERSAL_SUPPRESS_DEPRECATION`.
- `mixedprecision/tensor/CMakeLists.txt`: CG study dropped from the build (source left in place -- the mixed variants get redesigned in mp-iterative with a separate matvec/dot accumulator).
- `docs/migration/linear-algebra-extraction.md` documents where each piece moved.

[...]

Issue #1208, opened 2026-07-23
Phase 3: re-home BLAS experiments to mp-blas, mp-iterative, and new mp-numerics


Ravenwater:
Sub-issue of #1204. **Target repos: mp-blas, mp-iterative, NEW mp-numerics.** Depends on Phases 0-1.

Move the remaining Universal experiments onto MTL5 kernels + the quire shim, then delete each from Universal as it lands (the deletion itself is Phase 4).

## -> mp-blas (quire/fdp reproducibility + mixed-precision BLAS studies)
- [ ] `include/sw/blas/ext/fdp_blas.hpp` (fused matvec/matmul), `blas/mixed_precision.hpp`, reusable `blas/squeeze.hpp` bits, `applications/reproducibility/blas/*` (l1_fdp, l2_fmv, l3_fmm, lu, inverse, hilbert, randsvd, norms), `benchmark/*/blas/*`, `benchmark/accuracy/quantization/{mpdot,mpfma}.cpp` -> mp-blas, re-expressed via `accumulator_traits` + quire shim.

## -> mp-iterative (stationary + Krylov)
- [ ] `applications/accuracy/pde/{jacobi,gauss_seidel,sor,cg,laplace}.cpp` -> `mp-iterative/stationary` + `krylov`.
- [ ] The CG precision-variant study `mixedprecision/tensor/cg/*` (mv-dot x cmp-dot / cmp-fdp, mv-fdp x ...) -> `mp-iterative/krylov` (which already has `test_cg_quire`).

## -> new repo mp-numerics (decision #3)
- [ ] Scaffold `mp-numerics` from the mp-* template (namespace `sw::mp_numerics`, dual find_package/FetchContent).
- [ ] Migrate `mixedprecision/{roots,integration,interpolation,optimization,approximation}` (rpoly, simpson, chebyshev, etc.) -> mp-numerics.

## Acceptance
- [ ] Every experiment migrated and reproducing on MTL5; each target repo CI green.
- [ ] Migrated files ready to delete from Universal (Phase 4).

Part of #1204.

Ravenwater:
Phase 3 started -- first slice: **L2/L3 fused reproducibility studies -> mp-blas** (branch `feat/l2-l3-quire-studies`, pushed; PR pending a transient GitHub API hiccup).

[...]

Issue #1204, opened 2026-07-23
Epic: extract linear algebra from Universal -> MTL5 + mp-* (full removal)


Ravenwater:
## Context

Universal's BLAS (`include/sw/blas/`, namespace `sw::blas`) and its container types (`include/sw/numeric/containers/`, `sw::numeric::containers`) were built to study posit/quire dynamics. That functionality is now **superseded by MTL5** (`mtl::`, consumed as `MTL5::mtl5`), which already provides the full L1/L2/L3 surface, LAPACK-level factorizations (LU/QR/Cholesky/LDLt/SVD/eig), a complete Krylov suite (CG/BiCGStab/GMRES/MINRES/QMR/TFQMR/IDR-s), preconditioners, multigrid, sparse direct (native KLU), and a generic Universal-free `sparse/iterative_refine.hpp`. MTL5 replaces the quire with a 3-precision `mtl::math::accumulator_traits` policy.

So this is **not** a "port the BLAS into MTL5" effort (that would duplicate). It is: gap-fill the few utilities MTL5 lacks, re-home the *experiments* into the `mp-*` repos, and **fully remove** the linear-algebra code from Universal.

## Hard architectural constraint
**MTL5 must never depend on Universal.** All MTL5 + Universal coupling lives in the `mp-*` repos, via `include/mtl/math/quire_accumulator.hpp` (the quire -> `accumulator_traits` bridge). Anything touching quire/fdp/posit goes to `mp-*`, never MTL5.

## Target architecture (3 layers)
- `universal` -- number systems only (no linear algebra).
- `mtl5` (`MTL5::mtl5`) -- general LA kernels, generic over scalar, **Universal-free**.
- `mp-blas` / `mp-iterative` / `mp-ir` (new) / `mp-numerics` (new) / `mp-ode` / `mp-spice` -- mixed-precision experiments = MTL5 kernels + Universal types.

## Decisions (locked)
1. **End state: full removal** of `include/sw/blas/**`, `include/sw/numeric/containers/**`, `linalg/**`, `benchmark/*/blas`, BLAS-consuming `applications/*` + `mixedprecision/*`, and the `UNIVERSAL_BUILD_LINEAR_ALGEBRA_*` / `MIXEDPRECISION_SDK` CMake options -- after a one-release deprecation window.
[...]

PR #1212, opened 2026-07-24
chore(blas): Phase 4 start -- deprecate the linear-algebra layer; drop CG study from build

Ravenwater:
## Summary
Begins **Phase 4** (#1209) of the linear-algebra extraction (epic #1204): the **deprecation-window** step, ahead of removing Universal's BLAS + containers next release. Every experiment has now been re-homed to MTL5 + the mp-* repos (Phase 3, #1208), so Universal's linear-algebra layer can be wound down.

## Changes
- **`include/sw/blas/blas.hpp`**, **`include/sw/numeric/containers.hpp`** -- emit a `#pragma message` on include, pointing users to MTL5 + the mp-* repos and noting these headers will be removed. Informational only (does **not** fail `-Werror` on gcc/clang -- verified); silence with `-DUNIVERSAL_SUPPRESS_DEPRECATION`.
- **`mixedprecision/tensor/CMakeLists.txt`** -- drop the CG mixed-precision study (`./cg/*.cpp`) from the build. Its pure variants are migrated to mp-iterative (`applications/krylov/cg_precision`); the mixed variants will be **redesigned in mp-iterative** with a separate matvec/dot accumulator. The `./cg/` sources are **intentionally left in place** for reference during that redesign -- only the `compile_all()` is removed.
- **`docs/migration/linear-algebra-extraction.md`** -- where each piece moved (MTL5 / mp-blas / mp-iterative / mp-ir / mp-numerics) + the deprecation notice.

## Not in this PR (the removal step, after a deprecation release)
Full deletion of `blas/`, `numeric/containers/`, `linalg/`, `mixedprecision/`, `benchmark/*/blas`, the BLAS-consuming `applications/*`, and the `UNIVERSAL_BUILD_LINEAR_ALGEBRA_*` / `MIXEDPRECISION_SDK` CMake options.

## Validation
- `#include <blas/blas.hpp>` compiles under gcc **and** clang with `-Werror` (the pragma is a note, exit 0); `-DUNIVERSAL_SUPPRESS_DEPRECATION` silences it.
- ASCII guard passes.

Relates to #1209. Part of epic #1204.

## Summary by CodeRabbit

* **Documentation**
  * Added migration guidance for BLAS and numeric container functionality moving to MTL5 and related repositories, including replacement locations and deprecation timeline.

* **Deprecations**
  * Added compile-time deprecation warnings for legacy BLAS and numeric container headers.
  * Added a build flag to suppress these deprecation notices.

* **Build Changes**
  * Removed mixed-precision Krylov conjugate-gradient sources from the active build configuration (DNN build remains unchanged).

### DIFF
--- a/docs/migration/linear-algebra-extraction.md
+++ b/docs/migration/linear-algebra-extraction.md
@@ -1,0 +1,34 @@
+# Linear-algebra extraction (Universal -> MTL5 + mp-*)
+
+Universal's BLAS (`include/sw/blas/`) and container types
+(`include/sw/numeric/containers/`) were built to study posit/quire dynamics.
+That functionality is now superseded by **MTL5**
+(https://github.com/stillwater-sc/mtl5), a standalone C++20 linear-algebra
+library, and the mixed-precision *experiments* have moved to the **mp-*
+composition repos**. Tracking epic: universal#1204.
+
+## Where things moved
+
+| Was in Universal | Now |
+|------------------|-----|
+| `blas/` L1/L2/L3, solvers, generators, vmath, utes | MTL5 (`MTL5::mtl5`) -- already present; the few missing utilities (`linspace`, `magic`) were added |
+| `numeric/containers/` (matrix/vector/tensor) | MTL5 containers (`mtl::mat`, `mtl::vec`) |
+| dense LU iterative refinement (`ext/solvers/luir`, `squeeze`, `nbe`) | MTL5 `mtl::lu_iterative_refine` + `mtl::normwise_backward_error`; the posit/precision experiments in **mp-ir** |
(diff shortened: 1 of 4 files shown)


Ravenwater:
Addressed @coderabbitai in the latest commit:
- **[Minor] 120-col pragma lines** -- shortened both deprecation `#pragma message` strings so the lines fit (now 87 / 94 cols). I shortened rather than splitting into adjacent string literals, because `#pragma message` does not reliably concatenate adjacent literals on MSVC (which this ecosystem targets); the full detail (incl. `UNIVERSAL_SUPPRESS_DEPRECATION`) stays in the comment banner directly above each pragma. Still a note, not a warning, under gcc/clang `-Werror`.

Issue #1234, opened 2026-07-28
bug(interval): arithmetic does not round outward -- containment property does not hold


Ravenwater:
## Summary

`interval<Scalar>` performs its arithmetic with the ambient rounding mode instead of rounding **outward**, so the computed interval can be narrower than the true result range. The containment property — the Fundamental Theorem of Interval Arithmetic, and the only reason to use an interval type at all — does not hold.

This is a silent failure: no error, no warning, no compile diagnostic. Callers get an enclosure that looks rigorous and isn't.

Found against `173d8016` while scoping interval BLAS experiments in [mp-blas#12](https://github.com/stillwater-sc/mp-blas/issues/12).

## Reproduction

```cpp
#include <universal/number/interval/interval.hpp>
using I = sw::universal::interval<double>;

const double a = 0.1;
I p = I(a) * I(a);
double residual = std::fma(a, a, -(a * a));   // exact error of the product
```

```
a*a (rounded)  = 0.010000000000000001943
exact residual = -8.326672684688673745e-19    <- nonzero, so the product is INEXACT
interval       = [0.010000000000000001943, 0.010000000000000001943]   width = 0
```

The exact value of `a·a` lies strictly **below** the returned bound, and the interval has **zero width**. It does not contain the true product.

The same holds through composition — `mtl::dot` over 8 interval terms returns `[0.08, 0.08]`, width 0: an eight-term accumulation carrying no uncertainty whatsoever.

## Root cause

`interval_impl.hpp` uses plain arithmetic under round-to-nearest:

```cpp
interval& operator+=(const interval& rhs) noexcept {
    // [a,b] + [c,d] = [a+c, b+d]
    _lo = _lo + rhs._lo;      // must round DOWN
    _hi = _hi + rhs._hi;      // must round UP
    return *this;
}
```

The comment states the *mathematical* rule, which is correct over the reals. Realizing it in floating point requires directed rounding: `lo` toward −inf, `hi` toward +inf. Otherwise each endpoint can move inward by up to half an ulp per operation.

`grep -rn "nextafter\|nexttoward\|fesetround\|rounding\|round_up\|round_down\|inflate\|widen" include/sw/universal/number/interval/` returns **nothing** — the machinery was never present.

## Why the test suite doesn't catch it

`static/range/interval/arithmetic/{addition,subtraction,multiplication,division}.cpp` exist and pass, but every case uses small integers:

```cpp
Interval a(Scalar(1), Scalar(2));
[...]

PR #1248, opened 2026-07-29
feat(interval): Stage 2 EFT-based tight enclosures for native floats (#1247)

### LINKED ISSUE #1247: feat(interval): Stage 2 -- EFT-based tightness (widen only when the operation is inexact)
## Motivation
#1234 (PR #1246) restored the containment property for `interval<Scalar>` by rounding every inexact endpoint **outward** via `nextafter` (Stage 1). That is correct but **unconditional**: an operation that happened to be exact still widens by up to 1 ulp per endpoint. Over a composition the slack accumulates -- an 8-term interval dot carries ~8 ulp of avoidable width, and the exact-integer regression cases had to be converted from equality to containment assertions because their results now widen.

This is the optional Stage 2 from #1234: recover **1-ulp-optimal** enclosures by widening **only when the operation was actually inexact**, using error-free transformations (EFT).

## Goal
For each arithmetic operation, compute the rounded result `s` and the exact rounding error `e`; widen an endpoint only in the direction the true value lies:

- **`+` / `-` (TwoSum):** `s = fl(a + c)`, exact error `e`. If `e < 0` the true sum is below `s` -> widen the **lower** bound only; if `e > 0` -> widen the **upper** only; if `e == 0` -> widen neither.
- **`*` (TwoProduct):** `s = fl(a*c)`, exact error `e` via FMA (`fma(a, c, -s)`), or `quire_mul` for the Universal types that expose it. Same directional-widen logic.
- **`/`, `sqrt`, `sqr`:** derive the sign of the residual (e.g. `fma(s, d, -a)` for `s = fl(a/d)`) and widen the single implicated side.

Two payoffs: enclosures become as tight as the format allows, and the exact-operand regression tests (`static/range/interval/arithmetic/{addition,subtraction,multiplication,division}.cpp`) can assert **equality** again -- exact operations produce no widening.

## Notes / constraints
- Must stay **`Scalar`-generic** (like Stage 1's `nextafter` choice): `interval<posit<...>>` / `interval<cfloat<...>>` are targets. TwoSum is exact for any faithfully-rounded type; TwoProduct needs an FMA/`two_prod` -- the library already has EFT primitives to reuse: `internal/expansion/expansion_ops.hpp`, `number/elreal/block_eft.hpp`, `numerics/twoprod.hpp`, and the `dd`/`dd_cascade` two_sum/two_prod. `quire_mul` is available for posit/cfloat/lns/fixpnt (universal#1201/#1244/#1245).
- Keep the containment guarantee from #1234 as the invariant; Stage 2 only tightens, never narrows below the true range.

## Acceptance
- [ ] `+ - * / sqr sqrt` widen only when the operation is inexact; exact operations produce zero extra width.
- [ ] Containment still holds (reuse `static/range/interval/arithmetic/containment.cpp` fuzz).
- [ ] Tightness: enclosures within 1 ulp of optimal (strengthen the tightness bound in `containment.cpp`).
- [ ] The exact-operand arithmetic tests can revert to equality assertions and pass.
- [ ] Green on gcc + clang.

## References
- #1234 / PR #1246 -- Stage 1 (outward rounding, containment restored)
- Downstream: stillwater-sc/mp-blas#12 -- interval-BLAS enclosure-width characterization needs Stage 2 to measure *best-achievable* tightness.

Ravenwater:
## Summary
Stage 2 of the interval outward-rounding work (#1234 was Stage 1). Interval endpoints are now widened **only when an operation is actually inexact**, using error-free transformations, so exactly-representable operations stay exact while containment (the Fundamental Theorem of Interval Arithmetic) is preserved.

- `+`/`-`: Knuth **TwoSum** -> widen an endpoint iff the roundoff is nonzero
- `*`: Dekker-FMA **TwoProduct** per corner, then min/max of the tight enclosures
- `/`: FMA residual on the reciprocal endpoints, then the tight `*`
- `sqr`/`sqrt`: FMA residual (`s*s - x`) decides the widen direction

`interval<double>(2) + interval<double>(3)` is now `[5, 5]` instead of Stage-1's `[nextafter(5,-inf), nextafter(5,+inf)]`.

## Safety (the important part)
The tight path is gated by `interval_detail::interval_eft_exact`, enabled for **native `std::is_floating_point` types ONLY**. EFT roundoff terms are provably representable only for an IEEE-754-style format with round-to-nearest **and** gradual underflow (subnormals). The Universal fixed-size types do **not** qualify and keep the proven Stage-1 unconditional outward rounding:

- **cfloat without subnormals** (the default): the TwoSum error flushes to zero, so an inexact sum is mis-reported as exact -> containment silently breaks. Measured **132 breaks / 200k random pairs** for `cfloat<16,5>`.
- **posit**: tapered precision does not guarantee the roundoff is representable near the dynamic-range extremes.

The trait is **safe by default**: any unrecognized type resolves `false` and falls back to Stage-1, so containment can never be silently lost. Extending Stage 2 to `hasSubnormals` cfloat / verified posit configs is left as a follow-up.

## Also fixed
A sign bug in the EFT reciprocal for **negative-denominator** intervals: the widen direction depends on `sign(residual) * sign(denominator)`, not the residual sign alone. Verified with a 500k-interval division fuzz (124k negative denominators, 0 breaks).

## Changes
- `include/sw/universal/number/interval/interval_impl.hpp` - `interval_eft_exact` trait + `two_sum`/`two_prod`/`tight_lo`/`tight_hi` helpers; `if constexpr` EFT path in `+= -= *= /=`, `sqr`, `sqrt`, Stage-1 as the else branch
- `static/range/interval/arithmetic/{addition,subtraction,multiplication,division}.cpp` - trait-aware `enclosureOK`: exact equality for EFT types, containment for Stage-1 types
- `static/range/interval/arithmetic/containment.cpp` - division added to the fuzz (positive + negative denominators); new Stage-1-fallback containment fuzz for `cfloat<16,5>` and `posit<32,2>`; tightness bound tightened from 8 ulp to a genuine **1-ulp-optimal endpoint** check (fails under Stage-1 widening)

## Test Results
| interval_addition | OK | PASS | OK | PASS |
| interval_subtraction | OK | PASS | OK | PASS |
| interval_multiplication | OK | PASS | OK | PASS |
| interval_division | OK | PASS | OK | PASS |
| interval_containment | OK | PASS | OK | PASS |

Also reran interval_api / interval_conversion / interval_logic: PASS.

## Test plan
- [ ] Promote to ready when satisfied: `gh pr ready`

Resolves #1247

## Summary by CodeRabbit

* **Improvements**
  * Improved floating-point interval arithmetic to produce tighter, more accurate bounds for addition, subtraction, multiplication, division, squaring, and square-root operations.
  * Preserved existing handling for division by zero and invalid square-root inputs.
  * Improved accuracy across supported numeric formats, including fallback types.

* **Tests**
  * Expanded arithmetic validation to cover division and randomized interval scenarios.
  * Strengthened checks for enclosure correctness and near-optimal interval bounds.

### DIFF
--- a/include/sw/universal/number/interval/interval_impl.hpp
+++ b/include/sw/universal/number/interval/interval_impl.hpp
@@ -99,12 +99,20 @@ namespace interval_detail {
 	}
 	// enclose the exact value (s + roundoff) whose roundoff has sign 'roundoff':
 	// widen down only if the true value is below s, up only if above -- 1-ulp optimal.
+	// When the sum/product OVERFLOWED, s is +/-inf and roundoff is non-finite (NaN or
+	// +/-inf); the residual sign is then meaningless, so fall back to Stage-1
+	// unconditional outward rounding. round_down(+inf) is the largest finite value, which
+	// restores containment of the finite-but-unrepresentable true value (e.g. 1e308+1e308).
 	template<typename Scalar>
 	inline Scalar tight_lo(Scalar s, Scalar roundoff) noexcept {
+		using std::isfinite;
+		if (!isfinite(roundoff)) return round_down(s);
 		return (roundoff < Scalar(0)) ? round_down(s) : s;   // true value at or below s
 	}
 	template<typename Scalar>
 	inline Scalar tight_hi(Scalar s, Scalar roundoff) noexcept {
+		using std::isfinite;
+		if (!isfinite(roundoff)) return round_up(s);
 		return (roundoff > Scalar(0)) ? round_up(s) : s;     // true value at or above s
 	}
 }
@@ -253,14 +261,17 @@ public:
 		if constexpr (interval_detail::interval_eft_exact<Scalar>::value) {
 			using std::fma;
 			// residual r = fma(s, x, -1) = (s - 1/x) * x, so sign(s - 1/x) = sign(r)*sign(x).
-			// recipLo must stay <= 1/d: widen down iff s overestimates 1/d, i.e. r*d > 0.
+			// Compare signs directly rather than forming r*x: the product can underflow to
+			// zero for a tiny denominator (|r| ~ ulp(1), |x| subnormal), which would wrongly
+			// report "exact" and skip the widening.
 			Scalar d = rhs._hi, s = Scalar(1) / d;
 			Scalar r = fma(s, d, Scalar(-1));
-			recipLo = (r * d > Scalar(0)) ? interval_detail::round_down(s) : s;
-			// recipHi must stay >= 1/c: widen up iff s2 underestimates 1/c, i.e. r2*c < 0.
+			// recipLo must stay <= 1/d: widen down iff s overestimates 1/d, i.e. sign(r)==sign(d).
+			recipLo = (r != Scalar(0) && ((r > Scalar(0)) == (d > Scalar(0)))) ? interval_detail::round_down(s) : s;
 			Scalar c = rhs._lo, s2 = Scalar(1) / c;
 			Scalar r2 = fma(s2, c, Scalar(-1));
-			recipHi = (r2 * c < Scalar(0)) ? interval_detail::round_up(s2) : s2;
+			// recipHi must stay >= 1/c: widen up iff s2 underestimates 1/c, i.e. sign(r2)!=sign(c).
+			recipHi = (r2 != Scalar(0) && ((r2 > Scalar(0)) != (c > Scalar(0)))) ? interval_detail::round_up(s2) : s2;
 		}
 		else {
 			recipLo = interval_detail::round_down(Scalar(Scalar(1) / rhs._hi));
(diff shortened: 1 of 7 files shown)


Ravenwater:
Thanks @coderabbitai -- both actionable findings were real and are fixed in 1f91f46:

- **EFT overflow containment (Major)**: confirmed `interval<double>(1e308)+interval<double>(1e308)` produced `[+inf,+inf]` and did not contain the finite real `2e308`. Applied the `isfinite(roundoff)` guard to `tight_lo`/`tight_hi` (falls back to unconditional outward rounding), and added an overflow sum+product regression to `containment.cpp`.
- **Reciprocal sign test (Minor)**: switched `operator/=` from `r*x` to a direct `sign(r)` vs `sign(x)` comparison to avoid the product underflowing for a subnormal denominator.

Nitpicks also addressed: extracted the duplicated `enclosureOK` helper into `verification/interval_test_suite.hpp`, and fixed the stale "width" comment in the tightness test (it compares endpoints).

PR #1253, opened 2026-07-29
fix(interval): underflow-safe EFT product enclosure (#1252)

### LINKED ISSUE #1252: bug(interval): EFT product loses containment on underflow (subnormal products)
## Summary

The Stage-2 EFT tight path for `interval<Scalar>` multiplication (#1247) loses containment when a corner product **underflows** into the subnormal range. `interval<double>(a) * interval<double>(b)` returns a zero-width `[p, p]` that does **not** contain the true product, violating the Fundamental Theorem of Interval Arithmetic (#1234).

The overflow side of this was found and fixed during #1247 review (non-finite residual guard in `tight_lo`/`tight_hi`); the **underflow** side was missed. TwoProduct is exact only when the product does not under/overflow -- in the subnormal region the roundoff `fma(a,b,-p)` underflows below `denorm_min` and is lost, so the residual is reported as 0 (exact) when the product is actually inexact.

## Reproduction

```cpp
interval<double> X(1e-160), Y(1e-160);
auto R = X * Y;              // true product 1e-320 is subnormal
// R == [1e-320-ish, 1e-320-ish] (zero width) but does NOT contain the true product
```

Measured: **all 3,000,000** random subnormal-range products (operands ~2^-520) break containment. (Overflow, already fixed, and the normal range are fine.)

Same class affects `sqr` and, via reciprocal-then-multiply, `operator/=`.

## Root cause

`interval_detail::two_prod` + `tight_lo`/`tight_hi` trust `sign(e)` where `e = fma(a,b,-p)`. For a subnormal `p`, `e` cannot represent the roundoff (it is below `denorm_min`), so `e == 0` and the endpoint is not widened even though `a*b != p`.

## Fix (validated)

Replace the per-corner `two_prod` + tightening in the multiply path with an underflow/overflow-safe directed product enclosure `prod_enclose(a, b, lo, hi)`:
- normal product (`|p| >= min * 2^digits`, residual finite): use the direct EFT residual sign (1-ulp optimal, exact-preserving);
- overflow (`!isfinite(p)`): outward round;
- exact zero (`a==0 || b==0`): `[0, 0]`;
- subnormal / underflow: compute the **exact residual sign in a normalized (frexp) domain** -- `a=ma*2^ea, b=mb*2^eb`; `ma*mb = P+E` exact in the normal range; `resid = (P - ldexp(p,-(ea+eb))) + E` has the sign of `a*b - p` without underflow -- then widen only that side.

Standalone validation over normal / subnormal / deep-underflow (product -> 0) / exact-subnormal / with-zeros, using a true int128-mantissa exactness oracle: **0 containment breaks and 0 spurious widenings of exact products**.

Regression to add: subnormal-product containment cases in `static/range/interval/arithmetic/containment.cpp`.

Relates to #1247, #1234.

Ravenwater:
## Summary
Fixes a containment (Fundamental Theorem, #1234) violation in the Stage-2 EFT tight path (#1247): interval multiplication lost containment when a corner product **underflowed** into the subnormal range. TwoProduct's roundoff `fma(a,b,-p)` underflows below `denorm_min` and is lost, so `sign(e)` wrongly reports "exact" and the endpoint is not widened.

```cpp
interval<double>(1e-160) * interval<double>(1e-160)   // true product ~1e-320 (subnormal)
// was [p, p] (zero width) -- did NOT contain the true product
```
Measured: **100% of subnormal-range products** broke containment. The overflow side was already guarded (#1248, CodeRabbit); this fixes the underflow side, which the same EFT caveat produces.

## Fix
New `interval_detail::prod_enclose(a, b, lo, hi)` -- an underflow/overflow-safe directed product enclosure -- replaces the per-corner `two_prod` + `tight_lo`/`tight_hi` in `operator*=` and `sqr`:
- **overflow** (`!isfinite(p)`): outward round;
- **safely-normal** (`|p| >= min * 2^digits`, finite residual): direct EFT residual sign (1-ulp optimal, exact-preserving);
- **exact zero** (`a==0 || b==0`): `[0, 0]`;
- **subnormal/underflow**: exact residual sign in a normalized (`frexp`) domain -- `ma*mb = P+E` exact in the normal range, and `(P - ldexp(p,-(ea+eb))) + E` has the sign of `a*b - p` without underflow.

TwoSum is exact for finite (incl. subnormal) results, so `+=`/`-=` are unaffected (the sum overflow guard stays). The now-unused `two_prod` primitive was removed. (Corroboration: `ereal_impl.hpp` already `static_assert`s the same normal-double precondition for its Shewchuk EFTs.)

## Changes
- `include/sw/universal/number/interval/interval_impl.hpp` -- add `prod_enclose`; route `*=` and `sqr` through it; drop `two_prod`
- `static/range/interval/arithmetic/containment.cpp` -- subnormal + total-underflow + exact-zero product regressions, and a subnormal-product fuzz

## Validation
Standalone with a **true int128-mantissa exactness oracle**: 0 containment breaks and 0 spurious widenings of exact products across normal / subnormal / deep-underflow / exact-subnormal / zero, for both `double` and `float`.

## Test Results
| interval_addition | OK | PASS | OK | PASS |
| interval_subtraction | OK | PASS | OK | PASS |
| interval_multiplication | OK | PASS | OK | PASS |
| interval_division | OK | PASS | OK | PASS |
| interval_containment | OK | PASS | OK | PASS |

## Test plan

Resolves #1252
Relates to #1247

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved interval multiplication and squaring accuracy for very small and very large values.
  * Prevented underflow and overflow from producing incomplete or incorrect interval bounds.
  * Preserved exact zero results without unnecessary widening.

* **Tests**
  * Added coverage for subnormal products, total underflow, and exact-zero multiplication scenarios.

### DIFF
--- a/include/sw/universal/number/interval/interval_impl.hpp
+++ b/include/sw/universal/number/interval/interval_impl.hpp
@@ -90,19 +90,13 @@ namespace interval_detail {
 		Scalar bb = s - a;
 		e = (a - (s - bb)) + (b - bb);
 	}
-	// TwoProduct via FMA: p = fl(a*b), e = fma(a,b,-p) is the exact roundoff (RNE).
-	template<typename Scalar>
-	inline void two_prod(Scalar a, Scalar b, Scalar& p, Scalar& e) noexcept {
-		using std::fma;
-		p = a * b;
-		e = fma(a, b, -p);
-	}
 	// enclose the exact value (s + roundoff) whose roundoff has sign 'roundoff':
 	// widen down only if the true value is below s, up only if above -- 1-ulp optimal.
-	// When the sum/product OVERFLOWED, s is +/-inf and roundoff is non-finite (NaN or
-	// +/-inf); the residual sign is then meaningless, so fall back to Stage-1
-	// unconditional outward rounding. round_down(+inf) is the largest finite value, which
-	// restores containment of the finite-but-unrepresentable true value (e.g. 1e308+1e308).
+	// Used by the sum path (TwoSum is exact for finite results, even subnormal ones; the
+	// product path uses prod_enclose, which additionally handles underflow). When the sum
+	// OVERFLOWED, s is +/-inf and roundoff is non-finite (NaN); the residual sign is then
+	// meaningless, so fall back to unconditional outward rounding. round_down(+inf) is the
+	// largest finite value, restoring containment of the true value (e.g. 1e308+1e308).
 	template<typename Scalar>
 	inline Scalar tight_lo(Scalar s, Scalar roundoff) noexcept {
 		using std::isfinite;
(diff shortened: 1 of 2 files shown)


Ravenwater:
Thanks @coderabbitai -- valid catch. The -540 case rounded to zero (below `denorm_min`), duplicating the total-underflow test. Fixed in the latest commit: moved to -535 for a genuine inexact nonzero subnormal product, and added the exact-subnormal case (`2^-537 * 2^-537 == denorm_min`) asserting `[denorm_min, denorm_min]` -- which pins `prod_enclose`'s exact-preservation in the subnormal region. gcc + clang green.

Issue #1249, opened 2026-07-29
feat(interval): extend Stage 2 EFT tightness to sound fixed-size types (hasSubnormals cfloat, verified posit, bposit)


Ravenwater:
## Context

Stage 2 EFT-based interval tightness (#1247, PR #1248) is enabled for **native `std::is_floating_point` types only**. The tight path uses error-free transformations (Knuth TwoSum for `+/-`, Dekker-FMA TwoProduct for `*`, FMA residual for the reciprocal / `sqr` / `sqrt`) to widen an endpoint only when the operation is actually inexact, giving 1-ulp-optimal enclosures.

The gate is deliberately conservative because **EFTs preserve containment only when the roundoff term is guaranteed representable**, which requires an IEEE-754-style format with round-to-nearest-even **and** gradual underflow (subnormals). During #1247 this was shown to fail for the Universal fixed-size types:

- **cfloat without subnormals** (the default `hasSubnormals=false`): the TwoSum error term underflows to zero, so an inexact sum is mis-reported as exact and the endpoint is not widened. Measured **132 containment breaks / 200k random pairs** for `cfloat<16,5>` -- a silent violation of the Fundamental Theorem of Interval Arithmetic (#1234).
- **posit**: tapered precision means that in the extreme geometric-rounded regimes (where the fraction field shrinks toward zero bits near maxpos/minpos) the roundoff of a sum/product need not be representable, so TwoSum/TwoProduct are not unconditionally exact.

The trait `interval_detail::interval_eft_exact` in `include/sw/universal/number/interval/interval_impl.hpp` is safe-by-default (unknown type -> `false` -> proven Stage-1 unconditional outward rounding), and it is already structured to accept new `true_type` specializations. This issue tracks adding them, each gated on a proven representability precondition.

## Targets

1. **cfloat with subnormals** (`hasSubnormals=true`): add a partial specialization enabled only when the subnormal flag is set, and verify (fuzz vs a long-double / dyadic-exact oracle) that TwoSum/TwoProduct roundoff is representable across the normal + subnormal range. cfloat<16,5,...,true> showed 0 breaks in the #1247 probe -- promising, but must be proven, not sampled.
[...]

PR #1256, opened 2026-07-29
feat(interval): Stage-2 EFT tightness for cfloat with subnormals (#1255)

### LINKED ISSUE #1255: feat(interval): enable Stage-2 EFT tightness for cfloat with subnormals
Sub-task of #1249. Analysis there established that **cfloat with `hasSubnormals=true` is genuine IEEE-754** (RNE + gradual underflow): its TwoSum/TwoProduct are exact in the safe band exactly like native floats, and the only failures are the universal over/underflow EFT caveats (TwoSum intermediate overflow near maxpos; TwoProduct error underflow below denorm_min) -- the same caveats native `double` has, now guarded by the underflow-safe `prod_enclose` (#1252) and the `isfinite` overflow guard.

`cfloat<32,8>+sub` and `cfloat<64,11>+sub` show 0 TwoSum + 0 TwoProduct failures in their safe band (bit-for-bit IEEE); narrow configs just hit the guarded boundaries more often.

## Goal
Add `interval_detail::interval_eft_exact<cfloat<nbits,es,bt,true,sup,sat>> : std::true_type` (gated on the `hasSubnormals` flag) so `interval<cfloat<...,true>>` gets the Stage-2 tight enclosures, while `interval<cfloat<...,false>>` keeps the proven Stage-1 fallback.

## Design constraints (from the analysis)
1. **Gate strictly on `hasSubnormals=true`.** No-subnormals cfloat flushes the TwoSum error to zero and genuinely breaks containment (measured: 288 breaks / 300k for `cfloat<16,5>` without subnormals). The `false` case must stay Stage-1.
2. **prod_enclose needs a frexp-free underflow fallback for cfloat.** `frexp(cfloat<8,2>, &e)` does not normalize to [0.5,1) for narrow configs (returns 1.5, not 0.75, when the mantissa would be subnormal), so prod_enclose's normalized-domain residual path -- proven sound for native floats -- is unreliable for narrow cfloat. Use the frexp path only for `std::is_floating_point` types (`if constexpr`), and for cfloat fall back to sound 1-ulp outward rounding in the underflow region (preserving exact zeros). This is sound (contains the true subnormal product); it loses only exact-nonzero-subnormal tightness, which is negligible.
3. The safely-normal threshold `|p| >= min * 2^digits` already generalizes: for narrow configs the safe band shrinks (mostly fallback), for wide configs it behaves like native floats.

## Verification bar (soundness is containment-critical)
- Exhaustive/large containment fuzz over the FULL range of several configs: `cfloat<8,2>+sub`, `cfloat<10,4>+sub`, `cfloat<16,5>+sub`, `cfloat<32,8>+sub` -- 0 breaks vs a long-double oracle, for + - * / and sqr/sqrt.
- Exact-operand cases stay exact (enclosureOK equality) for wide configs.
- Add cfloat+sub to `containment.cpp` (EFT path) and keep the no-subnormals cfloat in the Stage-1-fallback fuzz.

Depends on #1252 (merged).

Ravenwater:
## Summary
Enables the Stage-2 EFT tight enclosures (#1247) for `interval<cfloat<...,true>>`. Per the #1249 analysis, cfloat with `hasSubnormals=true` is genuine IEEE-754 -- its TwoSum/TwoProduct are exact in the safe band exactly like native floats -- so it qualifies, with the over/underflow boundary caveats handled by the guards (`prod_enclose` #1252 + `isfinite`).

## Trait gate
```cpp
interval_eft_exact<cfloat<nbits,es,bt,true,sup,sat>> : bool_constant<(2*(nbits-es) <= 53)>
```
- **`hasSubnormals=true` required** -- no-subnormals cfloat flushes the TwoSum error and breaks containment (Stage-1 fallback stays).
- **`2*digits <= 53`** guarantees the exact product of two significands fits in a `double`, so `prod_enclose` verifies the rounding direction by exact double promotion -- robust where cfloat's own `frexp`/`ldexp` misbehave at narrow configs.

## Three robustness fixes (each needed for soundness)
1. `prod_enclose` branches on `is_floating_point`: native keeps the frexp normalized-domain path; cfloat uses double-verified direction.
2. `operator/=` reciprocal EFT residual gated to native floats; cfloat uses the Stage-1 reciprocal, then the tight `*=` recovers tightness.
3. `round_down`/`round_up` **clamp NaN to -/+inf**. `cfloat`'s `nextafter(maxneg,-inf)` returns NaN at the boundary; without the clamp an overflowing corner product produced a NaN that `std::min` swallowed -> a too-tight (unsound) lower bound. -/+inf is the correct directed value (also improves Stage-1 for cfloat).

## Verification
**0 containment breaks over 68M+ samples** (`+ - * / sqr sqrt`) across `cfloat<8,2>+sub`, `<10,4>+sub`, `<16,5>+sub`, `<32,8>+sub`, `<16,5>+sub+inf` -- wide (bit-for-bit IEEE) and pathologically narrow alike. Confirmed against a long-double oracle; Stage-1 comparison confirmed the narrow-config breaks were EFT-introduced (now fixed), not pre-existing.

## Changes
- `include/sw/universal/number/interval/interval_impl.hpp` -- trait specialization; `prod_enclose` cfloat branch; reciprocal gate; `round_down`/`round_up` NaN clamp
- `static/range/interval/arithmetic/containment.cpp` -- EFT cfloat<16,5>+sub and narrow cfloat<8,2>+sub containment fuzz; `VerifyContainmentFuzzT` scales range to the type + skips range-boundary enclosures

## Test Results
| Target | gcc | clang |
|--------|-----|-------|
| interval_addition / subtraction / multiplication / division | PASS | PASS |
| interval_containment | PASS | PASS |
| interval_api / conversion / logic | PASS | PASS |

## Test plan
- [ ] Fast CI passes

Resolves #1255
Relates to #1249, #1252

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved directed-rounding interval endpoints to correctly propagate genuine NaN inputs and avoid clamping NaNs to infinities.
  * Tightened interval multiplication/division behaviors, including more precise handling of underflow/subnormals and stricter gating for exact Stage-2 division refinement.
  * Prevented disjoint-interval “empty intersection” sentinels from being altered during interval construction.

* **Tests**
  * Expanded containment fuzz and added additional runs covering subnormal-enabled configurations, with smarter skipping for non-finite or out-of-range intermediate results.

### DIFF
--- a/include/sw/universal/number/interval/interval_impl.hpp
+++ b/include/sw/universal/number/interval/interval_impl.hpp
@@ -16,6 +16,8 @@
 #include <typeinfo>
 
 #include <universal/number/interval/exceptions.hpp>
+// forward decl of cfloat for the EFT-validity trait specialization (#1255)
+#include <universal/number/cfloat/cfloat_fwd.hpp>
 
 #ifndef INTERVAL_THROW_ARITHMETIC_EXCEPTION
 #define INTERVAL_THROW_ARITHMETIC_EXCEPTION 0
@@ -35,12 +37,18 @@ namespace interval_detail {
 	template<typename Scalar>
 	inline Scalar round_down(Scalar x) noexcept {   // toward -inf
 		using std::nextafter;
-		return nextafter(x, -std::numeric_limits<Scalar>::infinity());
+		Scalar ninf = -std::numeric_limits<Scalar>::infinity();
+		Scalar r = nextafter(x, ninf);
+		// some Universal types return NaN at the min boundary (nextafter(maxneg, -inf));
+		// the sound directed value there is -inf, so clamp (r != r is true only for NaN).
+		return (r == r) ? r : ninf;
 	}
 	template<typename Scalar>
 	inline Scalar round_up(Scalar x) noexcept {      // toward +inf
 		using std::nextafter;
-		return nextafter(x, std::numeric_limits<Scalar>::infinity());
+		Scalar pinf = std::numeric_limits<Scalar>::infinity();
+		Scalar r = nextafter(x, pinf);
+		return (r == r) ? r : pinf;   // clamp NaN at the max boundary to +inf
 	}
 	// widen a converted lower bound outward only when the conversion was inexact,
 	// so exactly-representable operands keep zero-width endpoints (interval<double>(2)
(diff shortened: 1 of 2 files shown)


Ravenwater:
Good catch @coderabbitai -- real regression. The clamp fired for genuine NaN inputs too, so `intersect()` of disjoint intervals (built as `interval(NaN,NaN)` via `enclose_lo/hi`) became `[-inf,+inf]` and `isnan()` returned false. Fixed with the `isnan(x)` guard (propagate genuine NaN, clamp only the finite-input boundary artifact), and added a regression asserting a disjoint intersection is NaN. Verified the cfloat boundary clamp still works (0 breaks / 68M) and gcc+clang green.

PR #1257, opened 2026-07-29
fix(cfloat): round_style is round_to_nearest, not toward_zero (#1254)

### LINKED ISSUE #1254: bug(cfloat): numeric_limits::round_style reports round_toward_zero but arithmetic is round-to-nearest-even
## Summary
`std::numeric_limits<cfloat<...>>::round_style` reports `round_toward_zero` (value `0`) for every configuration tested, but cfloat arithmetic actually rounds **to nearest, ties to even**. The trait is mislabeled; it should be `round_to_nearest` (value `1`).

## Evidence
```cpp
std::numeric_limits<cfloat<16,5,uint16_t,true,false,false>>::round_style   // == 0 (round_toward_zero)
std::numeric_limits<cfloat<32,8,uint32_t,true,false,false>>::round_style   // == 0
std::numeric_limits<cfloat<8,2,uint8_t,true,false,false>>::round_style     // == 0
```
But the arithmetic is demonstrably round-to-nearest-even:
- `cfloat<8,3>`: `0.75 + (-15.5)` computes `-15` -- RNE of the exact `-14.75` (tie -> even `-15`, not truncation toward zero which would give `-14.5`).
- `cfloat<8,2>`: `0.25 * 3.9375` computes `1.0` -- RNE of the exact `0.984375` (tie between subnormals `0.96875` and normal `1.0` -> even `1.0`).

Round-toward-zero would give the smaller-magnitude neighbor in both ties; it gives the RNE result instead.

## Impact
Downstream code that inspects `round_style` to decide whether error-free transformations (TwoSum/TwoProduct) are valid would wrongly conclude cfloat is not round-to-nearest. (Surfaced while analyzing EFT validity for `interval<cfloat>`, stillwater-sc/universal#1249.)

## Fix
Set `round_style = std::round_to_nearest` in cfloat's `numeric_limits` specialization. If any rounding mode other than RNE is ever actually implemented, the trait should track it; today the arithmetic is RNE, so the trait should say so.

Ravenwater:
## Summary
`std::numeric_limits<cfloat<...>>::round_style` reported `round_toward_zero`, but cfloat arithmetic rounds **to nearest, ties to even** (`cfloat_impl.hpp::round()` -- the classic guard/round/sticky round-to-even logic). `round_error()` was already `0.5` (the RNE value), corroborating the mislabel.

Downstream code that inspects `round_style` to decide whether error-free transformations are valid would wrongly conclude cfloat is not round-to-nearest. Surfaced while analyzing `interval<cfloat>` EFT validity (#1249 / #1255).

## Changes
- `include/sw/universal/number/cfloat/numeric_limits.hpp` -- `round_style = round_to_nearest`
- `static/float/cfloat/api/traits.cpp` -- regression asserting the label (`static_assert`) and the behavior: `0.25 * 3.9375 = 0.984375` is a tie between subnormal `0.96875` (odd) and normal `1.0` (even); RNE picks `1.0`, toward-zero would pick `0.96875`.

No production code branches on the value; cfloat api/logic/traits all pass.

## Test Results
| Target | gcc | clang |
|--------|-----|-------|
| cfloat_traits | PASS | PASS |
| cfloat_api | PASS | PASS |
| cfloat_logic | PASS | PASS |

Resolves #1254

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected cfloat numeric metadata to report round-to-nearest behavior.
  * Ensured halfway calculations use ties-to-even rounding consistently.

* **Tests**
  * Added coverage validating the reported rounding mode and a representative halfway-rounding result.

### DIFF
--- a/include/sw/universal/number/cfloat/numeric_limits.hpp
+++ b/include/sw/universal/number/cfloat/numeric_limits.hpp
@@ -72,7 +72,9 @@ public:
 	static constexpr bool is_modulo                = false;
 	static constexpr bool traps                    = false;
 	static constexpr bool tinyness_before          = false;
-	static constexpr float_round_style round_style = round_toward_zero;
+	// cfloat rounds to nearest, ties to even (see cfloat_impl.hpp::round(), the classic
+	// guard/round/sticky round-to-even logic), not toward zero (#1254).
+	static constexpr float_round_style round_style = round_to_nearest;
 };
 
 }
--- a/static/float/cfloat/api/traits.cpp
+++ b/static/float/cfloat/api/traits.cpp
@@ -119,6 +119,29 @@ try {
 		std::cout << "one more than that                           = " << scale(sp) + 1 << '\n';
 		std::cout << "std::numeric_limits<single>::max_exponent    = " << std::numeric_limits<single>::max_exponent << '\n';
 	}
+
+	// round_style must report round_to_nearest: cfloat rounds to nearest, ties to even
+	// (cfloat_impl.hpp::round()), NOT toward zero (#1254). Verify the label AND the behavior.
+	{
+		using Real = cfloat<8, 2, std::uint8_t, true, false, false>;   // with subnormals
+		static_assert(std::numeric_limits<Real>::round_style == std::round_to_nearest,
+			"cfloat round_style must be round_to_nearest (#1254)");
+		if (std::numeric_limits<Real>::round_style != std::round_to_nearest) {
+			++nrOfFailedTestCases;
+			std::cout << "FAIL: cfloat round_style is not round_to_nearest\n";
+		}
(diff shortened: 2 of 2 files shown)


Ravenwater:
Thanks @coderabbitai. The original tie is actually correct: for `cfloat<8,2>`, `min_normal = 1.0`, so `0.984375` is a **subnormal** value that is NOT exactly representable (`cfloat<8,2>(0.984375)` round-trips to `1.0`), and `0.96875`/`1.0` **are** adjacent (both step by `denorm_min = 0.03125`). So `0.25*3.9375` genuinely rounds to the even neighbor `1.0` -- the test passed for that reason. But you're right that a normal-range tie is clearer (it avoids the subnormal-boundary subtlety), so I'll add `1.5*1.09375 -> 1.625` alongside the subnormal case in a small follow-up. Both verified empirically.

PR #1259, opened 2026-07-29
fix(posit): to_native narrows once on return, fixing float subnormal flush

Ravenwater:
## Summary

`posit::to_native<Real>()` assigned each decoded factor to `Real` **before** forming the product:

```cpp
Real s = (_sign ? -1.0l : 1.0l);
Real r = _regime.value();            // long double -> Real, early
Real e = _exponent.value();          // long double -> Real, early
Real f = (1.0l + _fraction.value());
return s * r * e * f;
```

`positRegime`/`positExponent`/`positFraction::value()` all return `long double`, so those are four implicit narrowing conversions. MSVC reports each as **C4244 at every instantiation site**, which is how this surfaced — downstream builds are flooded with them (observed in stillwater-sc/mp-blas CI).

## The warnings were hiding a correctness bug

The conversions are lossy in **range**, not merely in precision. For a large `es` the regime factor alone can under/overflow `Real` even when the full product is perfectly representable. `posit<32,5>` and `posit<64,5>` → `float` hit exactly that: the regime underflowed `float` to 0, so the product came out 0 and representable **float subnormals were silently flushed to zero**.

```
posit<32,5> encoding 72   exact 2.22757e-40
                          correct float 0x00026cf5 (subnormal)
                          was          0x00000000   <-- wrong
posit<64,5> encoding ...  exact 1.62963e-44  ->  was 0
```

## Fix

Keep the whole product in `long double` — the width the components already return — and narrow **once**, explicitly, on return.

## Verification

Dumped every `posit`→`double` and `posit`→`float` conversion for 14 configurations before and after: **299,136 conversions**, exhaustive for `nbits <= 16` at `es = 0..3`, sampled for 32/64 bits.

- The **`double` column is bit-identical everywhere.**
- Exactly **448 float results changed**, all in `posit<32,5>` and `posit<64,5>`.
- Scoring each against the (exact) `double` conversion as ground truth: **old matched 0 of 448, new matches 448 of 448.** 446 of the 448 were zeros that should have been subnormals.

The result is bit-for-bit unchanged wherever the old code was already correct: regime and exponent are exact powers of two, so the only inexact factor is the fraction, and rounding it early then scaling by a power of two gives the same value as scaling first and rounding once. Only the range-clipping cases move.

## Regression test

Adds `static/tapered/posit/conversion/to_native_narrowing.cpp` (picked up by the existing `conversion/*.cpp` glob). It validates `to_native<Real>` against the `double` conversion of the same encoding — `double` has ample exponent range for these values, so it is exact ground truth.

Crucially it **fails on the pre-fix implementation** (751 mismatches for `posit<64,5>`, plus `posit<32,5>`) and passes after, while `posit<8,2>` / `<16,2>` / `<16,3>` / `<32,2>` pass either way — so it pins the fix without masking a regression.

## Also

Drops a stray duplicated semicolon on the `isnar()` line.

Draft because I have not run the full Universal regression suite locally — worth a CI pass before promoting. The mp-blas suite (10/10) and its `dot_accumulator_study` / `dot_characterization` outputs are bit-identical against this branch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved posit-to-native floating-point conversion accuracy, with correct handling of zero (including sign) and NaR (mapped to a quiet NaN).
  * Reduced narrowing-conversion discrepancies by making the conversion path more consistent.

* **Tests**
  * Strengthened regression coverage for `posit::to_native` narrowing by validating expected results only when a posit round-trips through `double` exactly, avoiding false failures from rounding differences.

### DIFF
--- a/include/sw/universal/number/posit/posit_impl.hpp
+++ b/include/sw/universal/number/posit/posit_impl.hpp
@@ -1200,18 +1200,30 @@ private:
 
 	template<typename Real>
 	Real to_native() const {
-		if (iszero())  return 0.0l;
-		if (isnar())   return std::numeric_limits<Real>::quiet_NaN();;
+		if (iszero())  return Real(0);
+		if (isnar())   return std::numeric_limits<Real>::quiet_NaN();
 		bool		     		_sign{ false };
 		positRegime<nbits, es, bt>   _regime;
 		positExponent<nbits, es, bt> _exponent;
 		positFraction<fbits, bt>     _fraction;
 		decode(_block, _sign, _regime, _exponent, _fraction);
-		Real s = (_sign ? -1.0l : 1.0l);
-		Real r = _regime.value();
-		Real e = _exponent.value();
-		Real f = (1.0l + _fraction.value());
-		return s * r * e * f;
+		// positRegime/positExponent/positFraction::value() all return long double,
+		// so keep the whole product in that width and narrow ONCE, explicitly, on
+		// return. Assigning each factor to Real instead (as this used to) forced
+		// four implicit long double -> Real conversions, which MSVC reports as
+		// C4244 narrowing warnings at every instantiation site.
+		//
+		// Bit-for-bit identical results: regime and exponent are exact powers of
+		// two, so the only inexact factor is the fraction, and rounding it early
+		// then scaling by a power of two gives the same value as scaling first and
+		// rounding once. Deferring the conversion additionally keeps the
+		// intermediates out of Real's range limits, which matters for
+		// configurations whose regime alone over/underflows Real even though the
+		// product does not.
+		const long double sign     = _sign ? -1.0l : 1.0l;
+		const long double scale    = _regime.value() * _exponent.value();
+		const long double fraction = 1.0l + _fraction.value();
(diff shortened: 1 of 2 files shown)


Ravenwater:
Both addressed in 45655c51:

1. **Comment misclaim** -- dropped "bit-for-bit identical results"; the change intentionally fixes ~448 encodings (0/inf -> correct value). Comment now says results are unchanged only where range is not a factor.
2. **Test oracle double-rounding** (good catch) -- confirmed your `posit<64,5> 0x4000000100000004` example: `double(p)` rounds it to the float midpoint, then `static_cast<float>` ties down to `1.0`, whereas the correct single rounding is `1+2^-23`. Fixed by using the double oracle **only when `double(p)` is exact** (skip when `posit(double(p)) != p`). This drops the high-precision near-1.0 cases the double oracle can't validate while retaining the range-underflow/subnormal cases the regression targets (few significand bits, double-exact): `posit<32,5>` keeps all 65536 sampled, `posit<64,5>` keeps 12261. A `long double` oracle would be tautological with the code under test, so I did not use it.

gcc + clang green.

Issue #1265, opened 2026-07-29
Epic: warning-clean headers under -Wconversion / C4244


Ravenwater:
## Goal

Make Universal's headers clean under `-Wconversion` / `-Wfloat-conversion` (GCC/Clang) and C4244 (MSVC), so downstream projects can build with narrowing diagnostics enabled and actually see their *own* problems.

## Why this matters downstream

Universal is header-only, so every warning in its headers is re-emitted in every consumer, once per template instantiation. A consumer build ([stillwater-sc/mp-blas](https://github.com/stillwater-sc/mp-blas)) currently reports **759 warning emissions** from Universal headers — but those come from only **28 unique source sites**. The multiplier is instantiation, not breadth.

The practical effect is that a consumer cannot keep `-Wconversion` on: real narrowing bugs in their own code are buried under hundreds of lines of dependency noise. The usual workaround is to mark the dependency's include directories `SYSTEM`, which suppresses the warnings — but that also suppresses genuine ones, so it is worth fixing the sites instead.

## Inventory

28 unique sites in 8 Universal headers (plus 1 in MTL5, tracked separately):

| file | sites | issue |
|---|---:|---|
| `internal/blockbinary/blockbinary.hpp` | 3 | #1260, #1261 |
| `internal/blocksignificand/blocksignificand.hpp` | 6 | #1260, #1261 |
| `number/posit/posit_impl.hpp` | 8 | #1264 + PR #1259 |
| `number/support/decimal.hpp` | 3 | #1263 |
| `number/cfloat/cfloat_impl.hpp` | 3 | #1260 |
| `number/lns/lns_impl.hpp` | 2 | #1260, #1264 |
| `number/posit/posit_exponent.hpp` | 1 | #1260 |
| `native/set_fields.hpp` | 1 | #1262 |

Note the site counts are *distinct (file, line, message)* triples; several lines warn under more than one target limb type (`uint8_t` / `uint16_t` / `uint32_t`), which is why e.g. `blocksignificand.hpp` has 6 for 2 lines.

## Sub-issues

- [ ] #1260 — limb-mask idiom `~(1ull << i)` implicitly narrows to `bt`, **copy-pasted across 5 files**. Largest single win; candidate for a shared helper so it stops reappearing.
[...]

PR #1272, opened 2026-07-30
fix: drop gratuitous ull in posit index arithmetic; lns scale() casts to int (#1264)

### LINKED ISSUE #1264: warnings: gratuitous `ull` in posit index arithmetic; lns scale() casts to long from an int function
## Summary

Two one-line narrowings that are independent of the mask-idiom issues (#1260, #1261).

### 1. `number/posit/posit_impl.hpp` 383–386: gratuitous `ull` in unsigned index arithmetic

```cpp
unsigned len   = 1 + std::max<unsigned>((nbits + 1ull), (2u + run + es));
bool blast     = pt_bits.test(len - nbits);
bool bafter    = pt_bits.test(len - nbits - 1ull);      // -Wconversion
bool bsticky   = pt_bits.anyAfter(len - nbits - 1ull);  // -Wconversion
```

`len` and `nbits` are `unsigned`, and `test`/`anyAfter` take `unsigned`, so the only thing `1ull` does is promote the expression to 64 bits so it can be narrowed straight back.

**Proposed fix** — use `1u` throughout:

```cpp
unsigned len   = 1 + std::max<unsigned>(nbits + 1u, 2u + run + es);
bool bafter    = pt_bits.test(len - nbits - 1u);
bool bsticky   = pt_bits.anyAfter(len - nbits - 1u);
```

Worth a glance while in there: `len - nbits - 1` is unsigned, so it wraps rather than going negative if `len < nbits + 1`. From `len = 1 + max(nbits + 1, ...)` we get `len >= nbits + 2`, so the expression is `>= 1` and it is safe today — but it is safe by a property of the line above rather than by construction, and the `ull` currently hides that by making the wrap astronomically large instead of merely large. Not asking for a redesign, just noting it while the lines are being touched.

### 2. `number/lns/lns_impl.hpp:445`: `scale()` returns `long` from an `int` function

```cpp
constexpr int scale() const noexcept {
    ExponentBlockBinary exp(_block);
    exp >>= rbits;
    return long(exp);      // -Wconversion: long -> int on LP64
}
```

`long` is 64-bit on LP64 and 32-bit on Windows, so this is a narrowing on Linux/macOS and a no-op on MSVC — which is why it shows up asymmetrically across the CI matrix. The cast looks like it was copied from a context with a wider return type.

**Proposed fix** — cast to the actual return type:

```cpp
return static_cast<int>(exp);
```

If a scale can genuinely exceed `int` range for large `nbits`/`rbits` configurations then the *return type* is what should widen, and the fix is to change the signature rather than the cast. `int` is what `scale()` returns across the other number systems, so I assume the cast is the mistake, not the signature — please confirm.

Ravenwater:
## Summary
Two independent one-line narrowings (warning-clean Epic #1265).

1. **posit** `posit_impl.hpp::convert_()`: the rounding bit-tests `test(len - nbits - 1ull)` / `anyAfter(len - nbits - 1ull)` (lines 385/386) promoted an unsigned index expression to 64 bits only to narrow it straight back to `unsigned` -> `-Wconversion`. `len`/`nbits` are `unsigned` and `test`/`anyAfter` take `unsigned`, so use `1u`. Also cleaned the gratuitous `nbits + 1ull` (line 383 -- constant that fits, didn't warn, but same idiom). `len >= nbits + 2` by construction, so `len - nbits - 1u >= 1` (no unsigned wrap).
2. **lns** `lns_impl.hpp::scale()`: returned `long(exp)` from an `int`-returning function -> `-Wconversion` (`long` is 64-bit on LP64, 32-bit on MSVC -- hence the asymmetric CI reports). **Confirmed** `int` is the `scale()` convention across the number systems (posit/cfloat return `int`), so the cast was the mistake -> `static_cast<int>`.

## Changes
- `include/sw/universal/number/posit/posit_impl.hpp` -- `1ull` -> `1u` (3 sites)
- `include/sw/universal/number/lns/lns_impl.hpp` -- `long(exp)` -> `static_cast<int>(exp)`

## Verification
The flagged lines are clean under `-Wconversion` on **gcc + clang**. Value-preserving.

| Target | gcc | clang |
|--------|-----|-------|
| posit_addition | PASS | PASS |
| posit_multiplication | PASS | PASS |
| lns_api | PASS | PASS |
| lns_conversion | PASS | PASS |

## Test plan
- [ ] Fast CI passes

Resolves #1264

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved numeric conversion reliability by making narrowing conversions explicit.
  * Corrected rounding and guard-bit calculations to prevent edge-case errors during posit number conversion.

### DIFF
--- a/include/sw/universal/number/lns/lns_impl.hpp
+++ b/include/sw/universal/number/lns/lns_impl.hpp
@@ -444,7 +444,7 @@ public:
 	constexpr int  scale()  const noexcept {
 		ExponentBlockBinary exp(_block);
 		exp >>= rbits;
-		return long(exp);
+		return static_cast<int>(exp);   // scale() returns int (matches posit/cfloat convention); long(exp) narrowed on LP64
 	}
 	constexpr blockbinary<nbits+2, std::uint32_t, BinaryNumberType::Unsigned> fraction() const noexcept {
 		blockbinary<nbits + 2, std::uint32_t, BinaryNumberType::Unsigned> bb{ 0 };
--- a/include/sw/universal/number/posit/posit_impl.hpp
+++ b/include/sw/universal/number/posit/posit_impl.hpp
@@ -380,10 +380,10 @@ constexpr inline posit<nbits, es, bt>& convert_(bool _sign, int _scale, const bl
 		pt_bits |= fraction;
 		pt_bits |= sticky_bit;
 
-		unsigned len = 1 + std::max<unsigned>((nbits + 1ull), (2u + run + es));
+		unsigned len = 1 + std::max<unsigned>((nbits + 1u), (2u + run + es));
 		bool blast = pt_bits.test(len - nbits);
-		bool bafter = pt_bits.test(len - nbits - 1ull);
-		bool bsticky = pt_bits.anyAfter(len - nbits - 1ull);
+		bool bafter = pt_bits.test(len - nbits - 1u);      // len >= nbits + 2, so this is >= 1 (no unsigned wrap)
+		bool bsticky = pt_bits.anyAfter(len - nbits - 1u); // 1u (not 1ull) keeps the index expression unsigned
 
 		bool rb = (blast & bafter) | (bafter & bsticky);
 


Ravenwater:
Thanks @coderabbitai -- I looked at both carefully; declining both, with evidence:

**posit_impl.hpp:386 (anyAfter / sticky bit) -- incorrect, declining.**
`anyAfter(bitIndex)` does NOT include its starting index. Its implementation (`blockbinary.hpp:810`) is:
```cpp
constexpr bool anyAfter(unsigned bitIndex) const noexcept {
    unsigned limit = (bitIndex < nbits) ? bitIndex : nbits;
    for (unsigned i = 0; i < limit; ++i) if (test(i)) return true;   // i < bitIndex -> excludes bitIndex
    return false;
}
```
So `anyAfter(len - nbits - 1)` already excludes the round bit `bafter` (at `len - nbits - 1`) and returns the OR of the strictly-lower bits -- the correct sticky. Empirical check: applying the proposed `- 2u` makes **posit_addition and posit_multiplication FAIL** (it drops a real sticky bit); the current `- 1u` passes both. So the change would introduce a rounding bug. (Also note: this line is pre-existing rounding logic; my PR only changed `1ull` -> `1u`.)

**lns_impl.hpp:447 (scale() return type) -- valid concern, but out of scope and pre-existing.**
`int` is the deliberate library-wide `scale()` convention (posit/cfloat `scale()` also return `int`), and the issue author explicitly confirmed it in #1264. Crucially, my change is **value-preserving**: `return long(exp)` in an `int`-returning function already narrowed to `int` on return, so the result is identical before and after -- the PR does not change the overflow behavior for wide configs, it only makes the cast match the return type and clears `-Wconversion`. Widening `scale()` (and doing so consistently across posit/cfloat/... to avoid divergence) is a separate library-wide design change, not part of this warning fix. Happy to file that as its own issue if desired.

PR #1276, opened 2026-07-31
docs(lns): document scale() int-truncation limit for wide configurations (#1274)

### LINKED ISSUE #1274: scale() returns int; truncates for wide configurations whose exponent exceeds int range
## Summary
`scale()` returns `int` across the number systems (posit, cfloat, lns, dbns, areal, dfloat, bfloat16, hfloat, e8m0, ...). For **wide configurations whose exponent range exceeds `int` (~2^31)**, `scale()` narrows/truncates, so the reported scale is wrong for those configs.

CodeRabbit flagged concrete examples on lns during PR #1272 review: `lns<64,11>`, `lns<96,32>`, `lns<128,15>` have exponent ranges that exceed `int`. The same applies to sufficiently wide posit/cfloat and any type whose exponent field can hold more than ~31 bits of magnitude.

```cpp
constexpr int scale() const noexcept {
    ExponentBlockBinary exp(_block);   // e.g. blockbinary<nbits-1> for lns
    exp >>= rbits;
    return static_cast<int>(exp);      // truncates when the exponent exceeds int range
}
```

## Why it is filed separately
This is **not** a regression -- it is a deliberate, library-wide convention (`int scale()` everywhere), and it long predates the warning fixes. #1264 only made the lns cast match the existing `int` return type (value-preserving); it did not change the truncation. Widening one type in isolation would diverge from the rest, so this needs a **library-wide decision**:

1. **Widen `scale()` to `int64_t`** consistently across all number systems (and audit callers/`ieee754_parameter`-style users), or
2. **Constrain / document** the supported `(nbits, es/rbits)` range so the exponent always fits `int` (a `static_assert` or a documented limit), or
3. Accept the current limit and document it explicitly.

## Impact
Low in practice today (the affected configs are unusual), but silently wrong when hit -- e.g. `lns<128,15>::scale()` cannot represent the true scale in an `int`.

Relates to #1264 / PR #1272 (CodeRabbit review). Part of Epic #1265 (correctness sweep).

Ravenwater:
## Summary
Per the #1274 decision (**document the limit**), add a doc note on `lns::scale()`.

`scale()` returns `int` by the library-wide convention (posit/cfloat/... all return `int`), and that convention is baked into the arithmetic engine (`blocktriple`/`blocksignificand` consume `int` scale) and ~499 call sites -- widening it is an Epic-scale rearchitecture. A `static_assert` was rejected because the wide configs are legitimately used (`lns<128,15>` inside the lns math library, `lns<64,11>`/`<128,15>` in perf tests).

`lns` is the one type whose exponent field (`nbits-rbits-1` bits) can exceed `int`, so `scale()` truncates for astronomically wide configs (`lns<64,11>` ~ 2^52, `lns<128,15>` ~ 2^111). The note records this and clarifies that lns **arithmetic is unaffected** (it operates on the full fixed-point exponent; only this metadata accessor saturates).

## Changes
- `include/sw/universal/number/lns/lns_impl.hpp` -- doc comment on `scale()`. **No behavior change.**

## Verification
| Target | gcc | clang |
|--------|-----|-------|
| lns_api | PASS | PASS |

## Test plan
- [ ] Fast CI passes

Resolves #1274

## Summary by CodeRabbit

* **Documentation**
  * Clarified that `lns::scale()` returns an `int` by library convention.
  * Documented that values may be truncated when the exponent exceeds the `int` range.

### DIFF
--- a/include/sw/universal/number/lns/lns_impl.hpp
+++ b/include/sw/universal/number/lns/lns_impl.hpp
@@ -444,11 +444,12 @@ public:
 	// scale(): the integer power-of-two exponent (the integer part of the lns exponent).
 	// Returns int by the library-wide scale() convention (posit/cfloat/... all return int).
 	// LIMITATION (#1274): the lns exponent field is nbits-rbits-1 bits wide, so for
-	// astronomically wide configurations (e.g. lns<64,11> ~ 2^52, lns<128,15> ~ 2^111)
-	// the scale exceeds int range and this accessor truncates. lns ARITHMETIC is
-	// unaffected -- it operates on the full fixed-point exponent, not on scale(); only this
-	// metadata accessor saturates. If a use needs the full scale for such configs, read the
-	// exponent field directly rather than through scale().
+	// astronomically wide configurations (e.g. lns<64,11> ~ 2^52, lns<128,15> ~ 2^111) the
+	// scale exceeds int range. operator int() keeps at most the low 64 bits and the C++20
+	// int64->int conversion then reduces modulo 2^32 (well-defined, but not the true scale).
+	// lns ARITHMETIC is unaffected -- it operates on the full fixed-point exponent, not on
+	// scale(); only this metadata accessor loses the high bits. There is currently no public
+	// full-width scale() alternative for such configurations.
 	constexpr int  scale()  const noexcept {
 		ExponentBlockBinary exp(_block);
 		exp >>= rbits;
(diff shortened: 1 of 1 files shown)


Ravenwater:
Both addressed in the latest commit, @coderabbitai -- valid on both:
- **saturates -> modulo wrap**: corrected to note `operator int()` keeps the low 64 bits and the C++20 `int64 -> int` conversion reduces modulo 2^32 (well-defined, not saturation).
- **exponent-field guidance**: removed the `read the exponent field directly` advice (`_block` is private; no public full-width accessor) and instead state that wide configs currently have no public full-width `scale()` alternative.

Issue #1236, opened 2026-07-28
bug(posit): numeric_limits has_infinity=true with infinity()==max()


Ravenwater:
## Summary

`std::numeric_limits<posit<nbits,es>>::has_infinity` reports **true**, and `infinity()` returns **maxpos** — the same value as `max()`. Posit has no infinity; it has a single NaR exception value. The sentinel is therefore indistinguishable from the largest finite value.

This is independent of #1234 (outward rounding) and is not fixed by it, but it produces a second, distinct containment violation in `interval<Scalar>`.

Found against `173d8016` while checking `interval<posit>` composition for [mp-blas#12](https://github.com/stillwater-sc/mp-blas/issues/12).

## Observed

```
has_infinity      = 1              <- posit has NaR, not infinity
infinity()        = 1.329228e+36
max()             = 1.329228e+36
infinity()==max()?  YES
has_quiet_NaN     = 1
quiet_NaN()       = nar   isnar=1  <- this one is correct
```

Systematic across widths:

| type | `has_infinity` | `infinity()` | `max()` | |
|---|---|---|---|---|
| `double` | 1 | `inf` | 1.79769e+308 | ok |
| `posit<8,2>` | 1 | 1.67772e+07 | 1.67772e+07 | :x: `infinity()==max()` |
| `posit<16,2>` | 1 | 7.20576e+16 | 7.20576e+16 | :x: |
| `posit<32,2>` | 1 | 1.32923e+36 | 1.32923e+36 | :x: |
| `posit<64,2>` | 1 | 4.52313e+74 | 4.52313e+74 | :x: |
| `lns<16,8>` | 0 | 1.83969e+19 | 1.83969e+19 | flag correct, but `infinity()` still returns a finite saturated value |

`quiet_NaN()` correctly returns NaR, so the NaN half of the specialization is right — only the infinity half is wrong.

## Why it matters: `interval<Scalar>` silently loses the unbounded result

`interval_impl.hpp`'s division path uses `std::numeric_limits<Scalar>::infinity()` to signal an unbounded result when the divisor interval contains zero:

```cpp
_lo = -std::numeric_limits<Scalar>::infinity();
_hi =  std::numeric_limits<Scalar>::infinity();
```

That is correct only when `Scalar` actually has an infinity. For `[1,4] / [-1,1]`, whose true result set over the reals is unbounded:

```
interval<double>: [-inf, inf]                      <- correct, entire
[...]

PR #1277, opened 2026-07-31
fix(interval): do not return a finite enclosure for an unbounded result

Ravenwater:
## Summary

Division by an interval containing zero has an **unbounded** exact result set, and the only enclosure of it is `[-inf, +inf]`. `interval<Scalar>` signalled that with `±numeric_limits<Scalar>::infinity()`, which assumes every `Scalar` **has** an infinity. Not every one does:

| Scalar | situation |
|---|---|
| `double`, `cfloat` | `has_infinity`, and `infinity() > max()` — can represent |
| `posit` | NaR instead of infinity; `numeric_limits<posit>::infinity()` returns **maxpos**, a finite value indistinguishable from ordinary data |
| `lns` | `has_infinity` is correctly `false`, but `interval` called `infinity()` without consulting the flag |

So `interval<posit<32,2>>` returned the **finite** `[-1.329e+36, 1.329e+36]` for a division whose exact result set is unbounded — an enclosure that excludes everything beyond maxpos while claiming to contain it.

Silently returning a too-narrow enclosure is the one failure mode interval arithmetic must never have. Every other error is conservative; this one is not.

## Fix

There is no correct value to return, so the operation now throws `interval_unrepresentable_unbounded`. That converts silent unsoundness into a loud, catchable failure.

```
interval<posit<32,2>>, [1,4]/[-1,1]:
  before   returned [-1.32923e+36, 1.32923e+36]    (finite, unsound)
  after    throws "unbounded result cannot be represented: Scalar has no infinity"
```

Adds `interval<Scalar>::can_represent_unbounded`, a compile-time query derived from what `infinity()` **actually is** rather than from the `has_infinity` claim, because the two disagree for posit:

```cpp
static constexpr bool can_represent_unbounded =
    numeric_limits<Scalar>::has_infinity &&
    (numeric_limits<Scalar>::infinity() > numeric_limits<Scalar>::max());
```

Callers can branch on it to skip element types that cannot express the answer, rather than discovering it at runtime.

## Scope

The guard fires **only** for a zero-containing divisor. Ordinary division is untouched for every Scalar (asserted in the test), as is every other operator.

## Testing

`static/range/interval/arithmetic/unbounded.cpp` covers both directions:

- `float` / `double` / `cfloat<32,8>` — must represent the unbounded result, **and** the returned interval must actually be unbounded (a finite return fails)
- `posit<16,2>` / `posit<32,2>` / `lns<32,16>` — must throw
- ordinary division for `double` and `posit<32,2>` — so the guard cannot creep

Verified against a downstream consumer: **stillwater-sc/mp-blas builds clean and all 16 of its tests pass** against this branch, including its interval BLAS suite (which exercises interval `dot`, `gemv`, `gemm`, `rot`, `nrm2`, and a naive interval `trsv`).

## Follow-up, deliberately not in this PR

A fuller fix would give `interval` an explicit **"entire"** state so the unbounded result is *representable* for every Scalar rather than merely refused. That means new state on a core class propagated through every operator — a larger change worth designing separately. This commit closes the unsoundness; the enhancement can restore the capability.

Related: stillwater-sc/mp-blas#12, where this was found while building an interval BLAS. Also stillwater-sc/universal#1236, which recorded the `numeric_limits<posit>` half of the diagnosis.

## Summary by CodeRabbit

* **New Features**
  * Improved interval division when the divisor contains zero.
  * Unbounded results are now represented using infinity when supported by the numeric type.
  * Clear exceptions are raised when unbounded results cannot be represented.

* **Tests**
  * Added validation across supported floating-point and specialized numeric types.
  * Confirmed ordinary interval division continues to work correctly.

### DIFF
--- a/include/sw/universal/number/interval/exceptions.hpp
+++ b/include/sw/universal/number/interval/exceptions.hpp
@@ -24,6 +24,14 @@ struct interval_negative_sqrt_arg : public interval_arithmetic_exception {
 	interval_negative_sqrt_arg() : interval_arithmetic_exception("negative sqrt argument") {}
 };
 
+// The exact result is unbounded, but Scalar has no infinity with which to
+// represent it. Thrown instead of silently returning a finite interval, which
+// would claim to enclose values it does not contain.
+struct interval_unrepresentable_unbounded : public interval_arithmetic_exception {
+	interval_unrepresentable_unbounded() : interval_arithmetic_exception(
+		"unbounded result cannot be represented: Scalar has no infinity") {}
+};
+
 ///////////////////////////////////////////////////////////////
 // internal implementation exceptions
 
--- a/include/sw/universal/number/interval/interval_impl.hpp
+++ b/include/sw/universal/number/interval/interval_impl.hpp
@@ -192,6 +192,23 @@ class interval {
 public:
 	using value_type = Scalar;
 
+	/// Can this interval represent an UNBOUNDED result?
+	///
+	/// Division by an interval containing zero has an unbounded exact result set,
+	/// and the only way to enclose it is +/-infinity. Some Scalars have no
+	/// infinity to offer: posit has NaR instead, and
+	/// numeric_limits<posit>::infinity() returns maxpos -- a finite value
+	/// indistinguishable from ordinary data. lns likewise reports
+	/// has_infinity = false.
+	///
+	/// The flag is derived from what infinity() actually IS, not from the
+	/// has_infinity claim, because the two disagree for posit. When this is false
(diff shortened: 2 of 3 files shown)


Ravenwater:
@coderabbitai review

Ravenwater:
Good catch @coderabbitai -- applied. The ordinary-division assertion now requires a FINITE enclosure strictly inside `+-max()` (`isfinite()` + strict bound checks), so it fails if ordinary division is ever routed through the zero-divisor path -- including the pre-fix posit case that returned `[-maxpos, maxpos]`. gcc + clang `interval_unbounded` PASS.

Issue #1297, opened 2026-08-11
takum: add the logarithmic takum as a distinct type over a shared codec


Ravenwater:
## Summary

Universal's `takum<nbits, rbits, bt>` implements the **linear** (floating-point)
takum. The takum literature also defines a **logarithmic** takum, which Universal
does not yet provide. This issue tracks implementing it.

Full assessment and rationale: [`docs/takum-design.md`](docs/takum-design.md)

## Background

- [arXiv:2404.18603](https://arxiv.org/abs/2404.18603) -- Definition 1, the
  **logarithmic** takum (this is the original paper).
- [arXiv:2408.10594](https://arxiv.org/abs/2408.10594) -- Definition 2, the
  **linear** takum (hardware codec paper).
- [libtakum](https://github.com/takum-arithmetic/libtakum) -- C99 reference
  implementation, which names the linear variant `takumN` and the logarithmic
  variant `takum_logN`. Universal's `takum<>` already matches that convention.

Both variants share the *identical* `(S, D, R, C, M)` bit layout, the identical
`r`/`c`/`m` decode, the identical two's-complement ordering, and the identical
zero/NaR encodings. They differ **only** in the value map:

| | logarithmic (Def. 1) | linear (Def. 2) |
|---|---|---|
| scale | `l := (-1)^S (c + m)` | `e := (-1)^S (c + S)` |
| value | `(-1)^S * sqrt(e)^l` | `[(1 - 3S) + f] * 2^e` |
| base | sqrt(e) ~= 1.6487 | 2 |
| dynamic range | ~ +/-[4.2e-56, 2.4e55] | ~ +/-[1.7e-77, 5.8e76] |
| mul/div/sqrt/pow | integer add/sub/shift on `l` | significand multiply |
| add/sub | Gaussian-logarithm tables | standard align-and-add |

## Decision

**Two distinct class templates over one shared codec** -- not a single type
parameterized by an encoding flag.

The codec is 100% shared, but it is the small part. The divergence lands above it:
arithmetic algorithms share zero code, the ~30 mathlib functions have unrelated
"good" implementations, every `numeric_limits` member differs, the quire story is
structurally different (a log takum cannot accumulate in the log domain), and the
constant tables are entirely disjoint. libtakum makes exactly this call: one
`src/codec.c`, eight distinct types, explicit cross-conversion entry points.

Universal precedent: `lns` and `cfloat` are already separate types despite both
being "sign + exponent + fraction" bit strings.

[...]

PR #1298, opened 2026-08-11
docs(takum): assess linear vs logarithmic takum, fix arXiv citation

Ravenwater:
## Summary

Adds `docs/takum-design.md`, an assessment of whether the linear and logarithmic
takum variants should be one parameterized type or two distinct types, and fixes an
incorrect arXiv citation in the `takum_impl.hpp` file header.

**No functional change.** Documentation and comments only.

## Background

The takum literature defines two variants:

- [arXiv:2404.18603](https://arxiv.org/abs/2404.18603) -- Definition 1, the
  **logarithmic** takum (the original paper).
- [arXiv:2408.10594](https://arxiv.org/abs/2408.10594) -- Definition 2, the
  **linear** takum (the hardware codec paper).

Universal's `takum<nbits, rbits, bt>` implements the linear one. This matches the
convention in [libtakum](https://github.com/takum-arithmetic/libtakum), the author's
C99 reference implementation, where `takumN` is linear and `takum_logN` is logarithmic
-- so no rename is needed.

Both variants share the *identical* `(S, D, R, C, M)` bit layout, the identical
`r`/`c`/`m` decode, the identical two's-complement ordering, and the identical
zero/NaR encodings. They differ **only** in the value map:

| | logarithmic (Def. 1) | linear (Def. 2) |
|---|---|---|
| scale | `l := (-1)^S (c + m)` | `e := (-1)^S (c + S)` |
| value | `(-1)^S * sqrt(e)^l` | `[(1 - 3S) + f] * 2^e` |
| base | sqrt(e) ~= 1.6487 | 2 |
| dynamic range | ~ +/-[4.2e-56, 2.4e55] | ~ +/-[1.7e-77, 5.8e76] |
| mul/div/sqrt/pow | integer add/sub/shift on `l` | significand multiply |
| add/sub | Gaussian-logarithm tables | standard align-and-add |

## Conclusion

**Two distinct class templates over one shared codec** -- not a single type
parameterized by an encoding flag.

The codec is 100% shared, but it is the small part. The divergence lands above it:
arithmetic algorithms share zero code, the ~30 mathlib functions have unrelated "good"
implementations, every `numeric_limits` member differs, the quire story is structurally
different (a logarithmic takum cannot accumulate in the log domain), and the constant
tables are disjoint. libtakum makes exactly this call: one `src/codec.c`, eight distinct
types, explicit cross-conversion entry points.

Universal precedent: `lns` and `cfloat` are already separate types despite both being
"sign + exponent + fraction" bit strings.

## Changes

- **`docs/takum-design.md`** (new) -- both specifications, the shared-encoding /
  divergence split, the recommendation with rationale, the proposed file structure, and
  references.
- **`include/sw/universal/number/takum/takum_impl.hpp`** -- the file header cited
  arXiv:2404.18603 as the source of the linear encoding. That paper defines the
  *logarithmic* takum; the linear takum is Definition 2 of arXiv:2408.10594. Corrected,
  with a note on the relationship between the two.
- **`include/sw/universal/number/takum/README.md`** -- was a stub with a dangling
  `reference:` line and nothing after it. Filled in.

## Testing

- `tools/scripts/check-ascii.sh` passes.
- No code paths touched; the only `.hpp` edit is inside a comment block.

## Follow-on

Implementation is tracked in #1297.

## Summary by CodeRabbit

- **Documentation**
  - Clarified that the existing `takum` format is linear and distinct from logarithmic takum.
  - Added design guidance for potentially supporting separate linear and logarithmic formats.
  - Updated references to relevant specifications, research papers, and implementation resources.
  - Documented the existing bit layout and value mapping without changing the public API.

### DIFF
--- a/docs/takum-design.md
+++ b/docs/takum-design.md
@@ -1,0 +1,157 @@
+# Takum: linear and logarithmic variants -- design assessment
+
+Status: assessment / proposal
+Date: 2026-08-11
+
+## Question
+
+The current Universal `takum<nbits, rbits, bt>` implements the *linear* takum. The
+takum literature defines both a linear and a logarithmic variant. Do they need to
+be two distinct types, or can they be parameterized into one type?
+
+## What the two specifications actually say
+
+The linear/logarithmic split is *not* in the original takum paper. Sequence:
+
+- **[arXiv:2404.18603](https://arxiv.org/abs/2404.18603)** -- "Beating Posits at Their
+  Own Game: Takum Arithmetic" (Hunhold, 2024). Defines takum as a **logarithmic**
+  tapered-precision format. This is Definition 1.
+- **[arXiv:2408.10594](https://arxiv.org/abs/2408.10594)** -- "Design and Implementation
+  of a Takum Arithmetic Hardware Codec". **Definition 2** introduces the **linear
+  takum**, and the paper builds a hardware codec for both.
+- **[libtakum](https://github.com/takum-arithmetic/libtakum)** -- the C99 reference
+  implementation by the same author. This is the authoritative tiebreaker on how the
+  author intends the split to be structured in software.
+
+### Shared encoding (identical in both variants)
+
+An n-bit string `T = (S, D, R, C, M)`, n >= 12:
+
+```
+[S:1][D:1][R:3][C:r][M:p]        r in {0..7},  p = n - r - 5
+
+r := uint(~R)  if D=0,          uint(R)        if D=1
+c := -2^(r+1)+1+uint(C) if D=0, 2^r-1+uint(C)  if D=1     in [-255, 254]
+m := 2^-p * uint(M)   in [0,1)
+
+zero: S=0, D=R=C=M=0
+NaR:  S=1, D=R=C=M=0
+```
+
+Two's-complement storage; negating the integer negates the value; integer ordering
+equals value ordering. All of this is byte-for-byte the same in both formats.
+
+### The only divergence: the value map
+
+| | logarithmic takum (Def. 1) | linear takum (Def. 2) |
+|---|---|---|
+| trailing field | mantissa `M`, `m` | fraction `F`, `f` |
+| scale | `l := (-1)^S (c + m)` | `e := (-1)^S (c + S)` |
+| value | `(-1)^S * sqrt(e)^l` | `[(1 - 3S) + f] * 2^e` |
+| base | sqrt(e) ~= 1.6487 | 2 |
+| dynamic range | ~ +/-[4.2e-56, 2.4e55] | ~ +/-[1.7e-77, 5.8e76] |
+| fraction bits | n-12 .. n-5 | n-12 .. n-5 (same) |
+| relative precision | constant across the whole range | sawtooth within each binade |
+| mul / div / sqrt / pow | integer add/sub/shift on `l` | significand multiply |
+| add / sub | Gaussian-logarithm tables | standard align-and-add |
+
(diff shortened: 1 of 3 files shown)


Ravenwater:
Thanks -- all seven findings verified against the extracted text of both papers and addressed in 9f4a6341.

**1. Definition numbering / provenance -- valid, and the core issue was worse than flagged.**

You were right that the original paper contains both variants; my framing ("the split is not in the original paper") was simply wrong. The exact numbering is:

| variant | 2404.18603 | 2408.10594 |
|---|---|---|
| logarithmic | **Definition 2** | Definition 1 |
| linear | **Definition 8**, Sec. 4.7 "Linear Takums" | Definition 2 |

(The mapping to "Definitions 2 and 20" in the comment is off -- the linear one is Definition 8. Algorithms 4 and 5 there give the float <-> linear-takum conversions, which is also what libtakum's `codec.c` means by "Algorithm 5 from the takum paper".)

All citations are now source-qualified across the design doc, the takum README, and the `takum_impl.hpp` header.

Reading the primary source for this also surfaced three things worth having:

- **2404.18603 Sec. 4.7 designates the *logarithmic* variant as the standard**: *"the logarithmic significand in Definition 2 is designated as the standard. It is incumbent upon implementations to explicitly specify whether they support 'linear takums' or 'logarithmic takums'. In the absence of such clarification, the logarithmic variant is to be presumed."* libtakum uses the opposite default (bare `takumN` = linear). My original text claimed Universal's naming "matches libtakum, no rename needed" without flagging that the spec says the opposite. Now called out explicitly in all three places.
- **Proposition 7** (bitwise inversion: integer increment of the bit string yields `1/x`) holds only for logarithmic takums -- an additional argument for two types that I had missed.
- **Propositions 3, 4, 6** (uniqueness, monotonicity, negation) hold for *both* variants, which is direct support for the shared-codec conclusion.

**2. Two's complement / NaR -- valid.** Narrowed to real encodings. Prop. 4 omits NaR from the ordering ("Conveniently omitted in Proposition 4 is the role of NaR..."), and Definition 7 sets it by separate convention; negating NaR is a no-op.

**3. Table notation and precision -- valid.** Euler's number is now `E` throughout so it no longer collides with the linear exponent `e`. You're right that `p in {n-12..n-5}` varies with the regime, so relative spacing is tapered in both -- the distinctive property is uniformity *within* a regime, not across the range. I replaced the hand-wave with the paper's actual result, Prop. 11: `lambda(p) = sqrt(E)^(2^-p-1) - 1` with `lambda(p) < (2/3) epsilon(p)`, i.e. the paper's own "at least two-thirds inferior". Log-variant trailing field renamed to "mantissa bits" per Definition 2.

**4. Exactness -- valid.** Now "exact before re-encoding rounding", with a note distinguishing removal of the transcendental step from the required output rounding.

**5. `numeric_limits::radix` -- valid and the most useful catch for the implementation.** `radix` is `static constexpr int` and cannot hold `sqrt(E)`. The doc now says to keep `radix` integral for `takum_log` and expose the value base as a separate takum-specific constant, plus a note that `min_exponent`/`max_exponent` bound the characteristic `c` rather than a power-of-two exponent.

**6. AVX10.2 title -- valid**, verified against the arXiv record and replaced with "Streamlining SIMD ISA Extensions with Takum Arithmetic: A Case Study on Intel AVX10.2".

**7. MD040 fences -- fixed**, both blocks tagged `text` (and the new one `cpp`).

The conclusion is unchanged and somewhat strengthened: two distinct types over one shared codec. `tools/scripts/check-ascii.sh` still passes; no functional change.

PR #1299, opened 2026-08-11
refactor(takum): extract the shared field codec into takum_codec.hpp

Ravenwater:
> **Stacked on #1298** (the design assessment). Base is `docs/takum-linear-vs-logarithmic-assessment`, so review that one first -- the diff here is against it. Both branches touch the `takum_impl.hpp` header comment.

## Summary

First implementation step of #1297: extract the field codec that the linear and logarithmic takum share, so `takum_log<>` can be added on top of it rather than duplicating it.

**No functional change.** Proven byte-for-byte -- see Verification.

## What moved

The two variants share an identical `(S, D, R, C, M)` bit layout and differ only in how the decoded `(c, m)` pair maps to a real number. New `takum_codec<nbits, rbits>` -- stateless, constexpr-clean, following the existing `zfpblock/zfp_codec.hpp` precedent -- owns:

| entry point | purpose |
|---|---|
| `layout_of(dr)` | DR / characteristic / mantissa geometry |
| `decode(magnitude)` | `-> {dr, c, M_bits, p}` |
| `encode_exact(c, M_bits)` | exact inverse of `decode`, integer fraction |
| `encode_rounded(c, m)` | round-to-nearest-even, carry propagation, saturation |

`encode_rounded` is the one that matters. The ~90 lines of rounding, carry-into-the-next-DR and saturation logic in `convert_ieee754` never referenced the value map, so both variants share it verbatim. `takum<>` now supplies only the linear value map: IEEE 754 fields -> `(c, m)` inbound, `(1 + f) * 2^c` outbound.

This mirrors libtakum's single `codec.c`, which decodes a linear takum by *"treat[ing] a linear takum as a logarithmic one: We obtain c and m from the logarithmic value l and compute f and e"*.

Two design points worth flagging for review:

- **Saturation is reported, not applied.** The codec has no storage, so it returns `takum_encode_status::{ok,overflow,underflow}` and the caller sets `maxpos`/`maxneg`/`zero` and owns the sign bit (invariant I4: the codec never sets it).
- **`encode_exact` (integer `M_bits`) is split from `encode_rounded` (double `m`)** so the `double` bottleneck stays confined to the float conversion path where it already existed, instead of silently capping the codec at `p <= 53` for `nbits` up to the allowed 64.

Also de-duplicates the field geometry, which was open-coded in **six** places: `characteristic()`, `convert_ieee754()`, `to_ieee754()`, `to_binary()`, and twice in `manipulators.hpp`.

## Public API unchanged

`overhead`, `dr_bits`, `nr_dr_values`, `max_r`, `r_mask`, `dr_field_mask`, `maxCharBits`, `dr_to_r`, `dr_to_c_bias`, `find_dr`, `min_characteristic`, `max_characteristic` are all re-exported from `takum<>`, so `manipulators.hpp`, `numeric_limits.hpp` and the `api/constexpr.cpp` `static_assert`s need no changes. A `static_assert` in the new test pins each re-export to its codec counterpart so they cannot drift.

## Verification

1. **Behavior equivalence against the pre-refactor tree.** Built a dumper against a worktree at the parent commit and against this branch: every representable bit pattern -> `double` (with `to_binary`, `scale()`, `iszero()`, `isnar()`), plus a 4200-point `double` -> bits sweep spanning and overflowing the dynamic range, across six configurations. **194,100 lines, byte-for-byte identical**, under both gcc and clang.
2. **New `static/tapered/takum/api/codec.cpp`** locks in representation invariants I1-I6 from `docs/takum-design.md`, exhaustively over every magnitude for six configurations -- including `takum<16,5>` and `takum<12,5>`, which exercise the degenerate `r > maxCharBits` path where `C` is stored truncated. Picked up automatically by the existing CMake glob.
3. **Mutation-tested** so the new suite isn't vacuous: dropping the truncation shift in `encode_exact` and perturbing the layout split are both caught (2,800 and 37,832 failures respectively).
4. Full takum suite passes under **gcc and clang**, no new warnings; `check-ascii.sh` passes.

## Next in #1297

`takum_log<nbits, rbits, bt>` on this codec, then its `numeric_limits`, traits, manipulators, mathlib and constants.

## Summary by CodeRabbit

* **New Features**
  * Improved Takum number encoding and decoding across supported formats.
  * Added more accurate rounding, overflow, and underflow handling.
  * Enhanced binary and formatted output consistency.

* **Bug Fixes**
  * Corrected field layout and characteristic/mantissa width calculations.
  * Preserved sign information during encoding and decoding.
  * Improved exact round-trip conversions across multiple Takum configurations.
  * Strengthened handling of wide fields, boundary values, and rounding carry behavior.

### DIFF
--- a/benchmark/performance/arithmetic/lns/log_add_algorithms.cpp
+++ b/benchmark/performance/arithmetic/lns/log_add_algorithms.cpp
@@ -296,7 +296,6 @@ try {
 
 	// Throughput iterations per config: scale down for the larger lns sizes
 	// so the benchmark completes in seconds rather than minutes.
-	constexpr std::size_t TPUT_TINY   = 1'000'000;
 	constexpr std::size_t TPUT_SMALL  =    75'000;
 	constexpr std::size_t TPUT_MEDIUM =    50'000;
 	constexpr std::size_t TPUT_LARGE  =    25'000;
--- a/docs-site/package-lock.json
+++ b/docs-site/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.0.1",
       "dependencies": {
         "@astrojs/starlight": "^0.41",
-        "astro": "^7.1",
+        "astro": "^7.2",
         "sharp": "^0.35"
       }
     },
@@ -202,35 +202,6 @@
       }
     },
     "node_modules/@astrojs/internal-helpers": {
-      "version": "0.10.1",
-      "resolved": "https://registry.npmjs.org/@astrojs/internal-helpers/-/internal-helpers-0.10.1.tgz",
-      "integrity": "sha512-5phcroT/vmOOrYuuAxtkbPixy5hePtlz9i8K4OeDv3dNK6/UQRuXPOSRTxIOBbUY5Sonw2UaxjbuVc43Mcir6Q==",
-      "license": "MIT",
-      "dependencies": {
-        "@types/hast": "^3.0.4",
-        "@types/mdast": "^4.0.4",
-        "js-yaml": "^4.1.1",
-        "picomatch": "^4.0.4",
-        "retext-smartypants": "^6.2.0",
-        "shiki": "^4.0.2",
-        "smol-toml": "^1.6.0",
-        "unified": "^11.0.5"
-      }
-    },
-    "node_modules/@astrojs/markdown-satteri": {
-      "version": "0.3.5",
-      "resolved": "https://registry.npmjs.org/@astrojs/markdown-satteri/-/markdown-satteri-0.3.5.tgz",
-      "integrity": "sha512-CvWVEFAbay7YO+i9SaqDJubipA5ckiVB89QWoMJ5XC0m5CtFg8JwZ7Kau6X9sYY7FZURH0w2l03ISH2jOS/RDQ==",
-      "license": "MIT",
-      "dependencies": {
-        "@astrojs/internal-helpers": "0.10.2",
-        "@astrojs/prism": "4.0.2",
-        "github-slugger": "^2.0.0",
-        "hast-util-from-html": "^2.0.3",
-        "satteri": "^0.9.1"
-      }
-    },
-    "node_modules/@astrojs/markdown-satteri/node_modules/@astrojs/internal-helpers": {
       "version": "0.10.2",
       "resolved": "https://registry.npmjs.org/@astrojs/internal-helpers/-/internal-helpers-0.10.2.tgz",
       "integrity": "sha512-yt7fMgPYqSM4Tmr+taTW6Per+hjJ8Pk6lA1PAcDyqzOt8HzJ6Kje5WzCxA2Sd+9wsUW7uhkLeoTMK0cXPwH9rQ==",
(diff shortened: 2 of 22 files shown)


Ravenwater:
## Codacy `ACTION_REQUIRED` -- diagnosed and fixed in 369de354

The gate is **"0 new issues"** and this PR introduced 5. Confirmed it was specific to this PR (every other recent PR, including the C++ change in #1293, is green) and reproduced locally:

```
cppcheck --enable=warning,style,performance,portability --std=c++20 -I include/sw \
  include/sw/universal/number/takum/takum_codec.hpp
```

All 5 were `unusedStructMember` on the new header:

| member | verdict |
|---|---|
| `decoded::M_bits`, `decoded::p` | false positive -- used, but only from other TUs |
| `encoded::magnitude`, `encoded::status` | false positive -- same |
| `decoded::dr` | **true positive** -- genuinely unused |

cppcheck cannot see through the class template to the instantiations in `takum_impl.hpp` and `api/codec.cpp`, so members read only from other translation units look dead to it.

**Fix.** Rather than suppress the four false positives, I gave the structs the accessors that belong on them anyway -- which also happens to use the members in-header:

- `decoded::fraction<Real>()` -- the trailing field as a real in `[0,1)`: the mantissa `m` of a logarithmic takum, the fraction `f` of a linear one. `to_ieee754` and the test now use it instead of open-coding the divide.
- `encoded::ok()` / `overflowed()` / `underflowed()` -- so call sites stop comparing against the enum directly.
- `encoded::sign_bit_clear()` -- invariant I4 as a named check; `api/codec.cpp` uses it for the I4 assertion.

For the true positive I took cppcheck's advice and **removed `decoded::dr`** -- no caller used it, and it is recoverable as `find_dr(c)` if a variant ever wants it. Shipping speculative API isn't worth defending.

Also dropped a stale `<cassert>` include left over from the contract-design pass; the codec documents its preconditions rather than asserting them, to stay constexpr-clean.

**cppcheck is now clean on both new files.** Re-verified after rebasing onto the post-#1298 `main`:

- 194,100-line behavior equivalence dump vs `origin/main` -- byte-for-byte identical under gcc and clang
- full takum suite passes on both compilers, 0 warnings
- `check-ascii.sh` passes

Ravenwater:
All six addressed in 89210d50. Verified each against the current code first -- one premise had gone stale.

**1. Duplicated DR reconstruction -> `dr_field()`** -- valid, and a fair catch: `(D ? (1u << rbits) : 0) + regime` is exactly `dr_field()`, and it was the last remaining copy of the geometry this PR set out to de-duplicate. Applied at all three sites (`to_binary`, `pretty_print`, `color_print`).

**2. Conditional-for-side-effects in the saturation branch** -- applied, plain `if/else`.

**3. Direct `<cstdint>` / `<type_traits>` in `api/codec.cpp`** -- applied.

**4. `takum_codec() = delete`** -- applied. The header claimed "never instantiated"; now the compiler enforces it.

**5. Assert `M_bits < 2^p` in `encode_exact`** -- applied. The substance is right and it's the one I'd rank highest: `pack()` ORs `M_bits` in without masking, so an oversized value corrupts the C and DR fields *silently*. Now loud in debug.

One correction: the premise that "the header already includes `<cassert>`, and that include is otherwise unused" was true of `3d2b6520`, the commit reviewed, but the Codacy fix in `369de354` had already removed it. Re-added now that the assert uses it.

I also checked the constexpr concern the review flagged implicitly: the assert does **not** disturb constant evaluation. The `constexpr auto e = C16::encode_exact(d.c, d.M_bits)` static assertion still compiles, and the suite passes both with asserts live and under `-DNDEBUG`.

**6. `codec<64,3>` coverage at p = 59** -- the technical claim checked out, so I verified it empirically before rewording anything:

```
dr=7,8  ->  p = 59                                    (codec<64,3>)
m * 2^59 exact:  1                                     <- no error from the scale step
59-bit M_bits through encode_exact:  round-trips exactly
59-bit M_bits through a double:      ...745 -> ...744  <- off by one
```

So you're right on both counts: no rounding error at large p, and a `p > 53` guard would be wrong. My "limited to p <= 53" was imprecise about *why* -- the limit is the input channel, not the arithmetic. A `double` carries at most 53 significant bits and so cannot express a wider trailing field in the first place. Reworded accordingly.

Added `VerifyWideTrailingField` over `codec<32,3>`, `<56,3>`, `<64,3>` and `<64,5>` -- non-exhaustive by necessity at 64 bits, but it exercises precisely the `encode_exact` / `encode_rounded` split that motivated having two entry points, which previously had no coverage above 16 bits.

**Re-verified:** 194,100-line equivalence dump vs `origin/main` still byte-for-byte identical under gcc and clang; full takum suite green on both compilers, 0 warnings; cppcheck clean on both new files; `check-ascii` passes.

Ravenwater:
Good catch, and a fair one -- fixed in ec6c2e6d.

You're right that `g.p <= 53` excluded exactly the case the test exists for: `VerifyWideTrailingField<64,3>` was never calling `encode_rounded` at its p = 59 layout. The test I added one commit ago was quietly vacuous above 53 bits.

The guard was there because my chosen fraction, `(2^(p-1) + 1)/2^p`, needs p significant bits and cannot survive a double past p = 53. Your suggestion fixes that by picking one that can.

I kept both fractions rather than swapping, since they demonstrate different things:

| fraction | representable | what it shows |
|---|---|---|
| `m = 0.5` | every p | reaches p = 59; the `m * 2^p` scale step is exact past 53 bits |
| `m = (2^(p-1)+1)/2^p` | p <= 53 only | agreement on a fraction that uses the full width |

Verified the new case is not vacuous in turn -- `codec<64,3>` now exercises 16 layouts, `encode_rounded` agreeing with `encode_exact` at each:

```
p=54 dr= 2 m=0.5 -> agree=1
...
p=59 dr= 7 m=0.5 -> agree=1
p=59 dr= 8 m=0.5 -> agree=1
cases exercised=16  max p reached=59
```

No production code touched by this commit. Full takum suite green on gcc and clang with 0 warnings, codec suite also passes under `-DNDEBUG`, 194,100-line equivalence dump vs `origin/main` unchanged, `check-ascii` passes.

Ravenwater:
## Coveralls `-0.02%` -- root-caused and fixed in bcc1c155

Not a threshold artifact. Measured with gcov locally instead of guessing, and it was pointing at a real gap.

`VerifyEncodeRounded` feeds back fractions it just decoded, so they are exactly representable and **no rounding ever happens**. That left the hardest logic in the codec untested:

| lines | path | status |
|---|---|---|
| 288-298 | `r > maxCharBits` truncation rounding + carry into next DR | was uncovered |
| 311-317 | M carry-out into C, and the DR bump | was uncovered |
| 147 | `find_dr` fallback | unreachable by construction |

These are exactly the lines lifted wholesale out of `convert_ieee754`, so they were the last thing that should have been going untested.

**Two new suites:**

- `VerifyEncodeCarry` -- fractions that round up and carry, ties that must go to even, carry past `max_characteristic` that must saturate, and `encode_exact`'s own overflow/underflow boundaries.
- `VerifyTruncatedCharacteristic` -- the `r > maxCharBits` path. Unreachable through a decode/encode round-trip, since `decode` already returns a quantized `c`, so it needs `c` values constructed directly.

Writing the second one **surfaced a case I had wrong**: a carry out of the *top* DR must saturate rather than advance, because there is no next DR. My first draft asserted success there and correctly failed against the code. Expectation fixed, codec unchanged -- `dr=63` returning `overflow` is right.

I also simplified `find_dr`: scanning down to `dr == 1` and returning 0 for the fall-through makes that return the legitimate answer for the lowest DR (`dr_to_c_bias(0) == min_characteristic()`) rather than an unreachable fallback. Same behavior, one less dead line. And dropped a stale inline comment still claiming "limited to p <= 53", which the top-of-file note already corrects.

**Result:** every executable line in `takum_codec.hpp` is covered by at least one instantiation, **85 of 85**. gcov's per-instantiation aggregate for the header goes 82.11% -> 87.92% (it stays under 100% because config-dependent branches cannot be taken by every instantiation -- e.g. `codec<64,3>` never truncates).

No production behavior change: 194,100-line equivalence dump vs `origin/main` still byte-for-byte identical under gcc and clang, full takum suite green on both compilers with 0 warnings, codec suite also passes under `-DNDEBUG`, cppcheck clean, `check-ascii` passes.

Ravenwater:
Both applied in ac60016c -- and both were regressions I introduced in the previous commit, so thanks for catching them.

**Misplaced comment.** Right: I inserted `fail_carry` *between* the explanatory comment and `VerifyTruncatedCharacteristic`, so it read as documenting the helper. Comment moved back onto the function it describes.

**Duplicate helper.** Also right -- `fail_carry` duplicated the `fail` lambda inside `VerifyEncodeCarry`. Both are now one shared `report_fail()`, declared once ahead of both users.

**Formatting.** Fixed the four-statement one-liner. While there I checked the whole file against `.clang-format` (`ColumnLimit: 120`, `TabWidth: 4`) and found 23 more over-long lines, nearly all in `main()`'s `ReportTestResult` table -- the flagged line was not the only one. Since `api/constexpr.cpp` and `api/api.cpp` in the same directory have zero lines over the limit, a new file alongside them should not either. All wrapped; the file is now clean at 120 columns counting tabs as 4.

No behavior change: same 27 test cases pass, 194,100-line equivalence dump vs `origin/main` byte-for-byte identical under gcc and clang, suite passes under `-DNDEBUG`, `check-ascii` passes.

For context on the previous round: coveralls is green again at 85.291%, and the coverage gap it flagged turned out to be real -- the carry and truncation paths in `encode_rounded` had no test at all. Every executable line in `takum_codec.hpp` is now covered, 85 of 85.

PR #1301, opened 2026-08-11
fix: resolve Apple Clang warnings without reducing test coverage

loren-osborn:
# Summary

This resolves the project-owned Apple Clang warnings found in a comprehensive Debug build with UNIVERSAL_BUILD_ALL=ON.

The initial build produced 126 warning instances across 19 source locations. The changes address the underlying causes rather than suppressing diagnostics.

Most were stale locals, constants, aliases, or computations, but reviewing the cleanup also identified two unused test constructions that represented implicit test intent. Those have been retained and converted into explicit assertions instead of simply being removed.

# Changes

* Remove unused S_bits constants from efloat IEEE conversion paths.
* Remove the unused sticky-bit computation from efloat::round(). The calculation did not participate in the function’s rounding decision; rint() retains its separate sticky-bit logic where required.
* Remove genuinely stale test, benchmark, and example locals and aliases.
* Scope firstTime to the LNS exception-enabled path where it is used.
* Turn the existing areal signed-integer constexpr construction into an explicit static_assert.
* Move INTEGER_THROW_ARITHMETIC_EXCEPTION before the related includes so the intended configuration is expressed without macro-redefinition ambiguity.
* Make previously implicit fixpnt runtime construction/arithmetic coverage explicit by comparing runtime Q32 operations with the corresponding constexpr results.
* Make the LNS near--1 cancellation setup explicit by checking that the selected nearby negative operand lies between -1 and 0.

No warning suppressions, -Wno-* options, diagnostic pragmas, or casts added merely to silence the compiler are included.

### Test-coverage review

A follow-up semantic audit specifically examined whether executing any of the removed “unused” expressions could itself have constituted useful test coverage.

### Two cases were retained and made explicit:

* Q32 runtime construction and arithmetic checks
* the exact -1 LNS construction used by the catastrophic-cancellation setup

Temporary mutation tests confirmed that deliberately breaking either new check causes its corresponding test to fail.

Other removed constructions were verified to be redundant or side-effect-free for their existing test context.

# Validation

Validated on Intel macOS with Apple Clang 21:

* Clean comprehensive Debug build with UNIVERSAL_ENABLE_TESTS=ON and UNIVERSAL_BUILD_ALL=ON
* 0 compiler warnings
* Focused affected tests: 14/14 passed
* Complete Debug CTest suite: 1033/1033 passed
* git diff --check passed
* Changed lines formatted with git clang-format

Focused GNU GCC 13.3 validation on Ubuntu 24.04 also passed without new warnings in the affected targets.

## AI-assisted development disclosure

OpenAI Codex performed the warning inventory, repository investigation, implementation, semantic audit, mutation checks, and most test execution.

ChatGPT reviewed the changes and raised the concern that some apparently unused test constructions might themselves represent test coverage. That prompted the additional audit in which Codex identified two such cases and converted their implicit coverage into explicit assertions.

I reviewed the resulting changes and validation before submitting them for maintainer review.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved fractional rounding behavior by simplifying guard-bit handling while preserving away-from-zero rounding.
  - Strengthened validation for catastrophic cancellation and special numeric values.

- **Tests**
  - Added runtime checks comparing fixed-point arithmetic with constexpr results.
  - Expanded compile-time coverage for signed integer ordering.
  - Improved division and conversion verification under configured arithmetic-exception modes.

- **Chores**
  - Removed unused benchmark, test, and conversion configuration elements.
  - Clarified numeric test setups without changing supported public interfaces.

### DIFF
--- a/benchmark/performance/arithmetic/lns/log_add_algorithms.cpp
+++ b/benchmark/performance/arithmetic/lns/log_add_algorithms.cpp
@@ -296,7 +296,6 @@ try {
 
 	// Throughput iterations per config: scale down for the larger lns sizes
 	// so the benchmark completes in seconds rather than minutes.
-	constexpr std::size_t TPUT_TINY   = 1'000'000;
 	constexpr std::size_t TPUT_SMALL  =    75'000;
 	constexpr std::size_t TPUT_MEDIUM =    50'000;
 	constexpr std::size_t TPUT_LARGE  =    25'000;
--- a/elastic/efloat/arithmetic/division.cpp
+++ b/elastic/efloat/arithmetic/division.cpp
@@ -89,8 +89,6 @@ namespace {
 		// --- IEEE 754 special values ---
 		double qnan = std::numeric_limits<double>::quiet_NaN();
 		double pinf = std::numeric_limits<double>::infinity();
-		double ninf = -pinf;
-
 		if (reportTestCases) std::cout << "  NaN / finite...\n";
 		{
 			efloat<16> a(1.0), n(qnan);
--- a/elastic/efloat/arithmetic/division.cpp
+++ b/elastic/efloat/arithmetic/division.cpp
@@ -88,8 +88,8 @@ namespace {
 
 		// --- IEEE 754 special values ---
 		double qnan = std::numeric_limits<double>::quiet_NaN();
-		double pinf = std::numeric_limits<double>::infinity();
-		if (reportTestCases) std::cout << "  NaN / finite...\n";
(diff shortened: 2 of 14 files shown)


loren-osborn:
@Ravenwater, Not in this commit, but now that CodeRabbitAI is enforcing `.clang-format`, we should probably format the whole codebase with a single sweeping, single-purpose commit.

Ravenwater (approved):
ship it

PR #1337, opened 2026-08-17
fix: return zero when dividing by infinity

### LINKED ISSUE #1327: x / inf returns NaN instead of zero in dd, qd, and all three cascade types
Found while fixing #1326; it is not caused by that change and it is not cascade-specific.

`x / inf` returns NaN in every multi-component type, where IEEE 754 says the result is zero:

```text
  dd          2/inf = nan
  qd          2/inf = nan
  dd_cascade  2/inf = nan
  td_cascade  2/inf = nan
  qd_cascade  2/inf = nan
```

`double` gets this right, so any code that swaps a `double` for one of these types changes behaviour
on this input.

## Cause

Division guards `rhs.iszero()` and the NaN cases explicitly, then falls through to the quotient-digit
loop. With an infinite divisor the first digit is `a[0] / inf = 0`, and the first residual is
`a - 0 * inf`, i.e. `a - NaN`, which poisons everything downstream.

`inf / x` is handled - `dd` guards a non-finite leading quotient explicitly, and the cascades now do
the same after #1326 - but an infinite *divisor* with a finite dividend produces a finite leading
quotient and is not caught.

## Fix

Guard a non-finite divisor before the digit loop, in the same place the zero-divisor case is already
handled, and return a correctly signed zero. Five types are affected, so the guard belongs wherever
the existing zero/NaN checks live in each.

## Note on scope

Worth fixing in all five together rather than only in the cascades: the value of the cascade types
is that they are drop-in replacements for the direct ones, and a divergence here would be worse than
the shared bug.

mikemikimike:
Closes #1327

## Background

Finite values divided by an infinite divisor returned NaN for `dd`, `qd`, and the three cascade types. IEEE 754 requires a correctly signed zero for this operation.

## Changes

- Add an infinite-divisor guard before each division algorithm.
- Return a zero with the sign determined by the dividend and divisor sign bits.
- Add a focused regression test covering positive and negative finite dividends across all five affected types.

## Compatibility

This only changes the previously incorrect finite-value / infinity case. Existing zero-divisor, NaN, and infinite-dividend handling remains unchanged.

## Verification

- Focused regression test with Docker GCC 13.3.0 — passed.
- `cmake -S /workspace/universal -B /workspace/build -G Ninja -DCMAKE_BUILD_TYPE=Release -DUNIVERSAL_ENABLE_TESTS=ON -DUNIVERSAL_BUILD_NUMBER_DOUBLE_DOUBLE=ON -DUNIVERSAL_BUILD_NUMBER_QUAD_DOUBLE=ON -DUNIVERSAL_BUILD_NUMBER_DD_CASCADE=ON -DUNIVERSAL_BUILD_NUMBER_TD_CASCADE=ON -DUNIVERSAL_BUILD_NUMBER_QD_CASCADE=ON -DUNIVERSAL_BUILD_REGRESSION_SANITY=ON` — passed.
- `cmake --build /workspace/build -j2` — passed.
- `ctest --test-dir /workspace/build --output-on-failure` — 113/113 passed.
- `git diff --check` — passed.

The Docker toolchain used GCC 13.3.0, Clang 18.1.3, CMake 3.28.3, and Ninja 1.11.1.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved division by infinity across extended-precision numeric types.
  * Finite values now produce correctly signed zero results, including proper handling of lower-precision components.
  * Infinity divided by infinity now returns NaN consistently.

* **Tests**
  * Added coverage for positive and negative operands, signed-zero results, and infinity-to-infinity division.


mikemikimike:
Addressed in commit `ee8a1de`: preserved `inf / inf` as NaN for all five types, made the `dd`/`qd` signed-zero branches constexpr-safe with the existing `is_constant_evaluated`/`bit_cast` pattern, and expanded coverage to negative-infinity divisors and infinity-over-infinity. Verification: focused regression passed under GCC 13.3.0 and Clang 18.1.3; Docker CMake build passed; CTest 113/113 passed.

mikemikimike:
Synced with the latest `upstream/main` via merge commit `eb01b21` (no force-push). Re-ran the focused regression under GCC 13.3.0 and Clang 18.1.3; both passed, and `git diff --check` passed. GitHub currently reports `BLOCKED` for merge-state checks, which is separate from the resolved out-of-date status.

mikemikimike:
Addressed the latest constexpr coverage review in commit `ebbaf9a`: added `static_assert` checks for all four signed-zero combinations for both `dd` and `qd`, using `std::bit_cast` on the high component. Made `dd::operator[]` constexpr to expose the existing constexpr division path. GCC/Clang focused tests and CTest 113/113 pass.

Ravenwater (approved):
lgtm

Ravenwater:
@mikemikimike thanks for submitting your first contribution. Looking forward to others. Any particular reason you picked the multi-component number system to work on?

mikemikimike:
I was just looking for a simple open-source PR for my coursework, but then I saw `inf / inf = NaN` and suddenly started taking it seriously.

mikemikimike:
I was just looking for a simple open-source PR for my coursework, but then I saw `inf / inf = NaN` and suddenly started taking it seriously.

Issue #1340, opened 2026-08-18
add_cascades<4> compresses after the fact: build the non-overlapping expansion directly


Ravenwater:
The last open framework item from the multi-component assessment
([Assessments / Multi-component Systems](https://stillwater-sc.github.io/universal/assessments/multi-component-cascade-vs-direct/), section 2).

`add_cascades<4>` sorts the eight input components by magnitude, accumulates them smallest-first with
`two_sum`, and emits the errors as they fall out. That sum is exact, but its components **overlap**,
so `compress_8to4` cannot keep the fourth one -- which is what made `qd_cascade` sqrt/exp/log 13-15
decimal digits worse than `qd` (#1317).

#1317 fixed the accuracy by adding a Shewchuk COMPRESS pass afterwards. It works, and it costs 46% of
the addition.

## The shape the algorithm should have

Compression is repairing an expansion that was built without the invariant in the first place. Both
inputs are already non-overlapping expansions in decreasing magnitude, so:

- **merging them by magnitude is linear** -- the bubble sort's 28 comparisons are redundant work on
  data that arrives sorted;
- **a single `two_sum` chain over the merged sequence produces a non-overlapping result directly**,
  which is Shewchuk's `fast_expansion_sum` (Geometric Predicates, Fig. 7). The compression pass then
  has nothing left to repair.

This is the same move that fixed multiplication in #1322/#1324 and trigonometry in #1318: where the
cascade improvises, adopt the proven schedule.

## Constraints

`add_cascades` has three consumers -- `operator+=`, the division residual, and the digit subtraction
in `to_decimal_string` -- and all three compress the 2N result back to N. The 2N expansion must
therefore stay **exact** (the division residual depends on it), not merely accurate to N components.
`fast_expansion_sum` is exact, so this is a reformulation rather than a trade.

## Acceptance

- the 8-term result still sums to a+b exactly, checked against exact dyadic arithmetic
- the result is non-overlapping, so `compress_8to4` keeps all four limbs
- `qd_cascade` addition accuracy no worse than today (the #1317 oracle)
- the 46% is recovered, or the measurement says why not

Ravenwater:
Follow-up filed as #1342: the merge is 27% faster under gcc and 7% slower under clang, shipped with the regression known. That issue carries the full measurement and the four formulations already tried.

Issue #1051, opened 2026-06-03
elreal: bfloat16 transcendentals via extended-precision intermediate host (defeatured from Phase 7)


Ravenwater:
**Parent epic: #923**
**Defeatured from: #931 (Phase 7)**

## Context
`#931` acceptance criterion 6 asks for the math suite parameterised over `{double, float, bfloat16}`. The cheap functions (`sqrt`, `hypot`, integer `pow`, the radical constants) ARE tested on bfloat16, but the **transcendentals are not**: a bfloat16 host has only a 7-8 bit mantissa (`k = numeric_limits<bfloat16>::digits = 8`), so a series expansion's blocks bottom out almost immediately and squaring a deep expansion underflows. The series/trig/exp functions are therefore scoped to wide hosts (`double`, `float`) today.

This is an **inherent limitation of using bfloat16 itself as the block host**, not a bug -- so it is being **explicitly defeatured from #931** and tracked here as its own feature.

## Idea
Support bfloat16 (and other narrow hosts, e.g. fp16) transcendentals by computing the series/reduction on an **extended-precision intermediate host** (e.g. evaluate at `FpType = double`, then convert/round the resulting `ZBCL` down to a `bfloat16`-hosted result), rather than running the expansion natively in bfloat16 blocks. This keeps the public `bfloat16`-hosted API while doing the lossy interior work where there is exponent/mantissa room.

## Open questions
- Where does the conversion live -- a host-rebinding utility on `ZBCL`? Depends on Phase 8 (#932) conversion.
- Is a `bfloat16` *result* even meaningful for an exact-lazy type, or is the real ask "compute in double, expose as bfloat16"?
- Same mechanism would unblock the deferred **fp16 host** (`half = cfloat<16,5>`), which additionally needs the div-floor rework.

## Acceptance criteria (future)
- [x] Decide the design (native narrow blocks vs extended-intermediate host).
- [ ] `sin/cos/tan/exp/log/...` available and tested on a `bfloat16` (and ideally fp16) host.
- [ ] Document the precision/semantics of a narrow-host transcendental result.

## Non-blocking
This does NOT block closing #931 -- #931 ships the transcendentals on `{double, float}` with bfloat16 transcendentals explicitly out of scope.

Ravenwater:
## The premise here needs revisiting: the limit is exponent range, not precision

[...]

PR #1360, opened 2026-08-20
docs(elreal): the narrow-host ceiling lifts -- 33 to 40 digits, not yet landable

Ravenwater:
## The shortcut works. The ceiling moved.

You asked me to try the `twoAdd` alignment shortcut. It works — with a one-bit correction — and **bfloat16 reached 40 decimal digits against a previous hard ceiling of 33**, with `double` bit-identical at 46/46.

It is not landable yet, and this PR is docs only. The tree is at baseline; a half-fixed invariant in core arithmetic would be worse than a known floor.

## Three attempts, each pinning a constraint

**1. `normalise()` alone — refuted.** Moving scale out of `v` spreads the `exp` fields, and `twoAdd`'s unconditional alignment (ldexp of `v` by the exp difference) then underflows the operand before the EFT sees it.

**2. The shortcut at threshold `k` — refuted, instructively.** It fails *on its own*, before `normalise()` is involved (`el_math_trigonometry`, 0-overlap at block 1). `twoSumRN` owes its callers the **round-to-nearest** sum plus residual, and `threeAdd` (Definition 4.2.1, from FCL.hs) is a fixed chain whose 0-overlap proof rests on that. Plain 0-overlap permits `b` up to just under `ulp(a)`:

| `b` | `RN(a+b) == a`? |
|---|---|
| quarter-ulp | yes |
| half-ulp | yes |
| 0.75 ulp | **no** |
| just under 1 ulp | **no** |

`{a,b}` is exact there but not round-to-nearest, and the chain doesn't survive the substitution.

**3. The shortcut at `k+1` — sound.** The extra bit puts `b` strictly below half an ulp, where RN cannot carry, so the pair really *is* the transform's own output. This is the non-overlapping vs **nonadjacent** distinction — the same one-bit margin, for the same reason, that governs the Shewchuk COMPRESS step in #1340.

## The result

`normalise()` in the divide/online-divide/online-multiply paths, the `k+1` shortcut, all three floors removed:

| depth | digits | blocks |
|---|---|---|
| 4 | 19 | 9 |
| 12 | 36 | 14 |
| **16** | **40** | 16 |
| 20 | 39 | 20 |
| 24 | *0-overlap assertion* | — |

The mechanism is confirmed: the floors were the cap.

## What keeps it out of the tree

**A 0-overlap violation at depth ≥ 24.** The non-monotonic 39 at depth 20 is the same fault showing early. Ruled out: `twoAdd`, the eager divide, the streaming divide, the streaming multiply — all normalised in the tried version. Some other producer still emits a pair closer than `k`. The `zbcl.hpp` assertion already names the offending pair, so instrumenting it to report the producer should localise it fast.

**A 23× slowdown**, 16s → 364s on the elreal suite. Part is real work — the series now refine to the working depth instead of stopping at a floor — but not 23× worth. `normalise()` calls `scale_of_v()` plus a wide `integer<256>` add on the hot path, and the shortcut adds two `exponent()` evaluations per `twoAdd`. `exponent()` is recomputed constantly and is the obvious thing to cache.

## Correction to this document

It previously said the approach was *"refuted by experiment"*. That was the verdict on one variant and the wrong conclusion to draw — the doc now reflects that the approach works and names precisely what is unfinished.

## Test Results

Tree is at baseline: **46/46 elreal**, gcc 13.3 and clang 18.1. `check-ascii.sh` clean.

## Test plan
- [ ] Fast CI passes

Relates to #1051 — still open, and now with a demonstrated path rather than a hypothesis.

### DIFF
--- a/docs/design/elreal-narrow-host-blocks.md
+++ b/docs/design/elreal-narrow-host-blocks.md
@@ -105,18 +105,24 @@ for `bfloat16` against 1.33 for `double` -- `double` is 6.8x more
 storage-efficient. A narrow host reaching high precision would need ~7.6x more
 blocks at ~0.9x the size each.
 
-## The proposal below was tried and does not work
+## Status: the approach works, and is not finished
 
-**Status of this section: refuted by experiment.** Kept because the reason is the
-useful part, and because it is the obvious idea -- anyone reading the diagnosis
-above will reach for it.
+**The ceiling has been lifted in a working tree: bfloat16 reached 40 decimal
+digits where it had been hard-capped at 33.** The diagnosis above is therefore
+confirmed end to end -- the floors were the cap, and removing them with the scale
+moved out of `v` does raise a narrow host past it.
 
-`block::normalise()` was implemented and behaves exactly as specified: it rescales
-`v` into `[1,2)`, folds the scale into `exp`, preserves the block's value and its
-combined exponent exactly (0 changes over 20,000 random blocks per host), and
-lifts subnormal `v` back to normal. The three floors were removed. The `double`
-suite stayed green -- 46/46, bit-identical -- which is what the gate below asked
-for.
+It is not landable yet. Two problems remain, both stated below. The record of
+what was tried, in order, is kept because each failure pins down a real
+constraint that the next attempt has to respect.
+
+### Attempt 1: normalise alone -- refuted
+
+`block::normalise()` behaves exactly as specified: it rescales `v` into `[1,2)`,
+folds the scale into `exp`, preserves the block's value and combined exponent
+exactly (0 changes over 20,000 random blocks per host), and lifts subnormal `v`
+back to normal. The three floors were removed. `double` stayed bit-identical at
+46/46, which is the gate below.
 
 `bfloat16` then aborted on the 0-overlap assertion.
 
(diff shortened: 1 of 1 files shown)


Ravenwater:
Superseded. This PR carried the intermediate findings from the #1051 investigation -- the ceiling lifting to 40 digits with a remaining 0-overlap violation and a 23x slowdown -- when the work was still incomplete.

All of it has since been overtaken and folded into the design note on main:

- The 0-overlap violation was traced to `twoDivZBCL`, not `addRec_step` as this PR reported, and fixed by normalising the divide's operands (#1361).
- The plateau and the slowdown both turned out to be the missing `twoSumRN` piece; with it the suite runs in 17s rather than 1469s (#1361).
- The gate that excluded `double` was itself wrong -- double was sitting on its own exponent wall at 307 digits -- and removing it took every host past its ceiling (#1362).

Final state: `elreal<double>` reaches 1598 digits at depth 96 and is linear at 16.0 digits/block with no ceiling; float reaches 319, bfloat16 146 and climbing. `half` remains capped and is tracked separately in #1363.

Closing rather than merging: the document on main supersedes this branch's version entirely, wrong turns included.

Issue #1444, opened 2026-09-13
refactor(headers): layer the static float-likes -- bfloat16, dbns, areal, hfloat, dfloat (#1334 group 5a)


Ravenwater:
Part of #1334, Phase 2 group 5. Covers the five **static float-like** types, which
share one recipe and one set of substrate re-points: `bfloat16`, `dbns`, `areal`,
`hfloat`, `dfloat`.

## Baseline, measured today

gcc 13.3, `-std=c++20`, `tools/scripts/measure-include-cost.sh`, on `main` at e873eded:

| type | preprocessed lines | transitive headers | I/O-family |
|---|---:|---:|---:|
| `bfloat16` | 93,035 | 377 | 5 |
| `dbns` | 83,961 | 406 | 5 |
| `areal` | 83,655 | 386 | 5 |
| `hfloat` | 78,566 | 399 | 5 |
| `dfloat` | 77,924 | 390 | 5 |

Acceptance is the epic's, unchanged: each new `core.hpp` at **0 of 5** I/O-family
headers, under 45,000 preprocessed lines where the type's dependency set allows it
(report the number reached rather than moving the bar), and every existing test
passing with the umbrella unchanged.

## Why this group is next, and why it is one issue

All five are **unblocked**. Every substrate they stand on is already layered and
measured at 0 I/O:

| substrate | lines | I/O | landed in |
|---|---:|---:|---|
| `native/ieee754_core.hpp` | 43,486 | 0 | Phase N |
| `native/manipulators_core.hpp` | 25,964 | 0 | #1421 |
| `internal/blockbinary/blockbinary.hpp` | 33,257 | 0 | Phase N |
| `number/quire/core.hpp` | 72,139 | 0 | #1435 |

`dbns` reaches `quire` by umbrella today and re-points to `quire/core.hpp`.

Three headers stay **out** of every core rather than being split -- this is the
pattern `cfloat/core.hpp` established, not a new decision:

| header | lines | I/O | why it stays out |
|---|---:|---:|---|
| `traits/arithmetic_traits.hpp` | 50,949 | 4 | three report-string builders; no traits in it |
| `utility/color_print.hpp` | 49,354 | 4 | already confined to `manipulators.hpp` tree-wide |
| `native/ieee754.hpp` | 62,519 | 4 | the umbrella; cores take `ieee754_core.hpp` |

## Where the I/O actually enters, per type

Traced with `g++ -H`. Two of the five need code to move; three need only includes
re-pointed:

[...]

Issue #1436, opened 2026-08-28
layered cores do not define their own *_THROW_ARITHMETIC_EXCEPTION guard


Ravenwater:
Every `number/T/core.hpp` produced by the #1334 rollout relies on its umbrella to define the behavioural switches that `T_impl.hpp` then tests in `#if` directives. A translation unit that includes only `core.hpp` — which is precisely what the epic tells callers they may do — leaves them undefined.

Measured with `-Wundef` on a core-only TU:

| type | undefined macro(s) |
|---|---|
| `edecimal` | `EDECIMAL_THROW_ARITHMETIC_EXCEPTION` |
| `fixpnt` | `FIXPNT_THROW_ARITHMETIC_EXCEPTION` |
| `posit` | `POSIT_THROW_ARITHMETIC_EXCEPTION` (5 sites) |
| `takum` | `TAKUM_THROW_ARITHMETIC_EXCEPTION` (2) |
| `rational` | `RATIONAL_THROW_ARITHMETIC_EXCEPTION` |
| `erational` | `ERATIONAL_…` **and** `EDECIMAL_…`, since it layers on edecimal's core |
| `qd` | `QD_THROW_ARITHMETIC_EXCEPTION` |

**Nothing is broken today.** `#if` on an undefined identifier evaluates to `0`, which is the intended default for all of these, and `-Wundef` is not in the project's compiler flags or in CI. The two real costs are:

1. it is a diagnostic for any consumer who does build with `-Wundef`, and
2. the core's configuration silently depends on an include the caller was told they do not need.

**Fix**, already applied to `quire` in #1435 as the pattern to copy: move the `#if !defined(X) / #define X 0 / #endif` block into `core.hpp`. The umbrella's identical guards then make including both a no-op — verified as 0 redefinition warnings through the umbrella.

This is mechanical and low-risk, but it touches seven merged subtrees, so it wants its own PR rather than riding along in a per-type split. Raised by CodeRabbit on #1435.

Part of #1334.

Ravenwater:
## Status: fixed in #1493 (c8875f37)

Every layered core now defines the behaviour switches it tests, so a translation unit that includes `core.hpp` alone sees the same values as one that includes the umbrella.

### What changed from the plan above

[...]

Issue #1389, opened 2026-08-24
headers: 148 headers use a std facility without including it (include-what-you-use sweep)


Ravenwater:
## Problem

148 headers under `include/sw/universal/` use a standard-library facility without including the header that declares it. They compile today only because something else in the include graph happens to pull it in.

This is not hypothetical fragility. It is the failure that broke the RISC-V cross build in #1386: `native/nonconstexpr/riscv_long_double.hpp` used `std::cerr` with no `<iostream>`, and it compiled everywhere until an unrelated include was removed. On x86 the `#if` selects a different file, so gcc and clang never saw it. Every header below is one include change away from the same thing, on whichever platform or standard library happens to expose it.

The header-layering epic #1334 makes include changes routine, so this moves from latent to load-bearing.

## Breakdown

| missing include | headers |
|---|---|
| `<string>` | 124 |
| `<sstream>` | 53 |
| `<iomanip>` | 36 |

Counting distinct headers rather than distinct misses: **148**.

## Already fixed

- #1388 fixed the 22 headers that PR touched, plus 60 that used `std::cout`/`cerr` with no `<iostream>`.
- The rest are untouched and listed below.

## Detection

```bash
python3 - <<'EOF'
import glob, re
need = {"<iomanip>": r'std::(setw|setprecision|setfill|setbase|quoted)\b',
        "<sstream>": r'std::(stringstream|ostringstream|istringstream)\b',
        "<string>":  r'std::string\b',
        "<iostream>": r'std::(cout|cerr|clog|cin)\b'}
for p in sorted(glob.glob("include/sw/universal/**/*.hpp", recursive=True)):
    s = open(p, encoding="utf-8", errors="replace").read()
    miss = [h for h, pat in need.items()
            if any(re.search(pat, l) and not l.strip().startswith("//") for l in s.split("\n"))
            and f"#include {h}" not in s]
    if miss: print(f"{','.join(miss):32s} {p}")
EOF
```

Worth folding into `tools/scripts/measure-include-cost.sh` or a CI check once the list is empty, so it cannot regrow.

## Remaining headers

```
<string>                         behavior/arithmetic.hpp
<iomanip>                        common/number_traits_reports.hpp
<string>                         energy/occurrence_energy.hpp
<sstream>,<string>               internal/abstract/triple.hpp
<string>                         internal/bitblock/bitblock.hpp
<sstream>                        internal/floatcascade/floatcascade.hpp
[...]

Issue #1334, opened 2026-08-17
Epic: layer the number-system headers so a compute kernel does not pay for I/O


Ravenwater:
# Epic: layer the number-system headers so a compute kernel does not pay for I/O

Opening work item for V5. The original proposal is preserved in the first comment; this body replaces it with a measured baseline, a checkable target, and a phase plan.

## Problem, measured

Including one number system and nothing else, gcc 13.3 `-std=c++20`, on `v4.10.1`:

| type | preprocessed lines | transitive headers | I/O-family headers | `-fsyntax-only` |
|---|---|---|---|---|
| `posit` | 114,754 | 494 | 5 | 0.452 s |
| `cfloat` | 115,034 | 486 | 5 | 0.443 s |
| `lns` | 110,397 | 477 | 5 | 0.414 s |
| `fixpnt` | 109,425 | 464 | 5 | 0.418 s |
| `integer` | 103,543 | 425 | 5 | 0.386 s |
| `dd` | 107,684 | 459 | 5 | 0.472 s |
| `qd` | 109,694 | 460 | 5 | 0.485 s |

For scale, on the same compiler:

| reference point | lines |
|---|---|
| `<iostream>` alone | 43,165 |
| `<sstream>` alone | 44,318 |
| `<iomanip>` alone | 49,285 |
| `<string>` alone | 31,825 |
| `<cstdint> + <type_traits> + <limits> + <cmath> + <concepts>` | 24,977 |
| ...the same plus `<string>` | 42,091 |

**A scalar compute kernel that never prints anything currently pays roughly 110k preprocessed lines and ~480 headers to get one arithmetic type.** In a simulation sweep that is multiplied across every translation unit, and across the ~1,900 test binaries this repo builds.

## Root cause: two distinct costs, not one

The original proposal named one cause -- member methods like `to_binary()` and `color_print()` coupling the class template to `<string>`/`<sstream>`/`<iostream>`. That is real and confirmed:

```
include/sw/universal/number/posit/posit_impl.hpp:11:  #include <iostream>
include/sw/universal/number/posit/posit_impl.hpp:12:  #include <iomanip>
include/sw/universal/number/posit/posit_impl.hpp:15:  #include <sstream>
include/sw/universal/number/cfloat/cfloat_impl.hpp:27: #include <sstream>
```

[...]

PR #1447, opened 2026-09-13
refactor(hfloat): split hfloat into core / manipulators / iostream (#1334)

Ravenwater:
Third type of #1444 (Phase 2 group 5a of #1334).

## Acceptance numbers

gcc 13.3, `-std=c++20`:

| include | lines | headers | I/O-family |
|---|---:|---:|---:|
| `hfloat.hpp` before | 78,566 | 399 | 5 |
| `hfloat.hpp` after | 77,199 | 385 | 5 |
| **`core.hpp`** | **59,518** | **304** | **0** |
| target | under 45,000 | -- | 0 |

Zero I/O is met. Under-45,000 is not, and this reports the number reached rather than
moving the bar: `utility/decimal_to_binary.hpp`, which `parse()` converts through, is
**57,369 lines on its own** (already at 0 I/O), so this core sits within 2,200 lines of
its single largest dependency. Same shape as dfloat (#1446) -- the ceiling belongs to the
substrate, not the type.

## `str()` had to be reimplemented -- this PR's one real change

hfloat's `str()` is core surface (`operator<<` formats through it), but unlike dfloat's,
which concatenates, it built its result with a stringstream:

```cpp
std::stringstream ss;
if (nrDigits > 0) ss << std::setprecision(nrDigits);
ss << d;
```

That one function was the **only** reason this core would have needed `<sstream>`.

`std::ostream`'s `defaultfloat` presentation of a `double` *is* printf's `%g` at the
stream's precision (which defaults to 6), so `str()` now reaches the same formatter
through `snprintf("%.*g")` -- the `<cstdio>` idiom Phase 0 adopted -- with a two-pass call
so arbitrary precision still works, and a clamp on the `size_t` -> `int` narrowing.

**Verified, not assumed.** A standalone harness compared the two formulations over
**604,742 (value, precision) pairs** -- the full double exponent range from 1e-320 to
1e308, 40 mantissas per decade, precisions 0 through 60, plus every boundary value
(`min`, `max`, `denorm_min`, signed zeros) -- with **zero mismatches on gcc and clang**.
The differential below then compares **184,176 further `str()` renderings in situ**.

One observable difference remains, and it is deliberate: a stringstream carries the
**global locale** while `snprintf` uses the **C locale**, so a program that calls
`std::locale::global` with a non-C numeric facet would previously have seen a localised
decimal separator from `str()` and now will not. Nothing in the tree sets one, and for a
library rendering a hexadecimal float's value the C locale is the defensible answer --
but it is a real difference and should be a conscious call, not a silent one.

## What else moved

- **`to_binary()`, `to_hex()`, `to_native()`** format through a stringstream ->
  `manipulators.hpp`.
- **`operator<<`, `operator>>`** -> `iostream.hpp`, which includes **only `core.hpp`**:
  `operator<<` formats through `str()` and `operator>>` hands its token to `parse()`, both
  core, so there is no dependency on `manipulators.hpp` at all.
- **`parse()` stays in the core.** It converts through `utility/decimal_to_binary.hpp`,
  already at 0 I/O, so it is arithmetic surface -- the same call dfloat's `parse()` got,
  and the opposite of cfloat's and bfloat16's, which run an `istringstream`.
- `native/ieee754.hpp` -> `native/ieee754_core.hpp`.
- `HFLOAT_ENABLE_LITERALS`, `HFLOAT_THROW_ARITHMETIC_EXCEPTION` and `HFLOAT_EXCEPT` now
  default in `hfloat_impl.hpp` beside the code they govern (#1436 applied up front).
- `manipulators.hpp` and `attributes.hpp` had **no includes at all** -- they reached
  `std::stringstream`, `std::setw`, `Color`, `type_tag` and `operator<<` entirely through
  the umbrella's include order. Both now name what they use (#1389).

## Verification

- **Layer gate** (#1390): `core.hpp` alone, and `core.hpp` + `manipulators.hpp`, each
  compiled, linked and run under `-Wall -Wpedantic` on gcc 13.3 and clang.
- **Self-containment**: core, manipulators, iostream, attributes and the umbrella each
  compile standalone with 0 errors on both compilers.
- **12 targets built and run green on both compilers**: the 10 hfloat tests plus `ucalc`
  and `ucalc_regression`.
- **Differential against `main`**, from a worktree: hfp32 and hfp64 across *every*
  exponent code against 24 fraction patterns; all three formats across 155 decades x 4
  mantissas driven through `parse()`; each row rendered by `to_binary` (both nibble
  settings), `to_hex`, `to_native`, `color_print`, **`str()` at 23 precisions**,
  `operator<<` under six stream states, plus negation, square and double.

  **7,896,511 bytes of stdout and all 30 stderr diagnostics byte-identical** -- old vs new
  on gcc, old vs new on clang, *and* gcc vs clang.

  The cross-compiler check earned its place: the first version of the probe built hfp128
  encodings with `uint64_t(e) << 120`, which is UB, and gcc and clang duly disagreed on
  5,890 lines. That was the probe, not the headers -- `setbits(uint64_t)` cannot reach
  hfp128's exponent field at bits 120..126 at all -- and the wide format is now driven
  through `parse()` instead.

## Summary by CodeRabbit

* **New Features**
  * Added stream input and output support for hfloat values, including preservation of formatting settings.
  * Added binary, hexadecimal, and native-format string representations for hfloat values.
  * Improved parsing error reporting when reading hfloat values from streams.

* **Refactor**
  * Organized hfloat functionality into more focused components, improving compatibility for arithmetic-only use cases.
  * Improved string conversion reliability and efficiency while preserving existing interfaces.

### DIFF
--- a/include/sw/universal/number/hfloat/attributes.hpp
+++ b/include/sw/universal/number/hfloat/attributes.hpp
@@ -5,6 +5,16 @@
 // SPDX-License-Identifier: MIT
 //
 // This file is part of the universal numbers project, which is released under an MIT Open Source license.
+//
+// Part of the text layer (#1334): hfloat_range() streams hfloat values through a
+// stringstream, so this header sits above iostream.hpp and is not included by core.hpp.
+// Self-contained: <string>, <sstream> and <iomanip> were all used but never named.
+#include <string>
+#include <sstream>
+#include <iomanip>   // std::setw
+#include <universal/number/hfloat/core.hpp>
+#include <universal/number/hfloat/manipulators.hpp>   // type_tag
+#include <universal/number/hfloat/iostream.hpp>       // operator<< on hfloat, used by hfloat_range
 
 namespace sw { namespace universal {
 
--- a/include/sw/universal/number/hfloat/core.hpp
+++ b/include/sw/universal/number/hfloat/core.hpp
@@ -1,0 +1,44 @@
+#pragma once
+// core.hpp: the hfloat arithmetic core -- no I/O, no text
+//
+// Copyright (C) 2017 Stillwater Supercomputing, Inc.
+// SPDX-License-Identifier: MIT
+//
+// This file is part of the universal numbers project, which is released under an MIT Open Source license.
+//
+// Layer 1 of the hfloat headers (#1334, Phase 2 group 5a, #1444). Storage, arithmetic,
+// comparison, numeric_limits and traits -- everything a compute kernel needs and
+// nothing that turns an hfloat into text.
+//
+//     #include <universal/number/hfloat/hfloat.hpp>   // everything, as before
+//     #include <universal/number/hfloat/core.hpp>     // arithmetic only
+//
+// hfloat.hpp includes this plus manipulators.hpp, iostream.hpp and attributes.hpp, so
+// existing code is unaffected.
+//
+// WHAT STAYS HERE, and why:
+//   - str() is core surface: operator<< formats through it. It used to build its result
+//     with a stringstream, which was the single reason this core would have needed
+//     <sstream>; it now reaches the same formatter through snprintf("%.*g"). See the
+//     note on str() in hfloat_impl.hpp.
+//   - parse() converts through utility/decimal_to_binary.hpp, which is already at zero
+//     I/O-family headers, so it is arithmetic surface rather than text -- the same call
+//     dfloat's parse() got (#1446), and the opposite of cfloat's and bfloat16's, whose
+//     parse() runs an istringstream.
+//
+// native/ieee754.hpp was re-pointed to native/ieee754_core.hpp, its stream-free half.
+//
+// NOTE: traits/arithmetic_traits.hpp and common/number_traits_reports.hpp are
+// deliberately NOT included -- they build report strings and pull <sstream>/<iomanip>.
+// The umbrella provides them.
+#include <universal/utility/architecture.hpp>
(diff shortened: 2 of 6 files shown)


Ravenwater:
On CodeRabbit's `to_hex()` finding: **valid, and pre-existing -- filed as #1448 rather than fixed here.**

Verified both halves of that claim:

- The `to_hex()` body in this PR is **byte-identical to `main`** (`git show main:...hfloat_impl.hpp` vs the relocated copy diffs clean). This PR moves the function from `hfloat_impl.hpp` to `manipulators.hpp` and changes nothing inside it.
- The behaviour reproduces on `main`: a zero renders `+0x0.000000 * 16^-64`, and because hfloat has no single zero encoding -- any exponent field with an all-zero fraction is a zero -- the sweep also produced `+0x0.000000 * 16^-63`. Two values that compare equal print different exponents.

It is not fixed here because it would change output, and #1334 requires its rollout PRs to be bit-identical -- that is what makes them safely mechanical. This PR's verification is a differential showing **7,896,511 bytes matching `main` byte for byte**; applying the suggested one-liner would break that guarantee and bury a behaviour change inside a refactor. Same call made for #1434 and #1436.

#1448 carries the fix, including the parts that make it more than a one-liner: `to_native()` delegates to `to_hex()` so both change together, the signed-zero rendering wants deciding at the same time, and the commit needs a `BEHAVIOUR-CHANGE:` trailer.

PR #1452, opened 2026-09-13
refactor(areal): split areal into core / manipulators / iostream / debug (#1334)

Ravenwater:
Last type of #1444 (Phase 2 group 5a of #1334). areal had the group's largest `std::cout`
population -- 51 sites -- and almost all of them were already correctly guarded.

## Acceptance numbers

| include | lines | headers | I/O-family |
|---|---:|---:|---:|
| `areal.hpp` before | 83,655 | 386 | 5 |
| `areal.hpp` after | 82,620 | 372 | 5 |
| **`core.hpp`** | **73,412** | **345** | **0** |
| target | under 45,000 | -- | 0 |

Zero I/O met. Under-45,000 not met, reporting the number reached:
`internal/blocktriple/blocktriple.hpp` is **69,758 lines on its own** (already 0 I/O), so
this core is blocktriple plus about 3,650. Same ceiling dbns hit (#1449).

## The 51 cout sites were not the problem -- the include was

**49 of them already sat behind areal's `TRACE_CONVERSION` switch**, which was already
externally settable: the defect #1387 found elsewhere is genuinely absent here. What kept
four I/O-family headers in every areal translation unit was that the
`#include <iostream>` sat **outside** the guard. It now sits inside it, with the code that
needs it -- the same move cfloat's trace includes got in #1417/#1418.

The remaining two:

- **`constexprClassParameters()`**, a member function with 19 `cout` statements, is
  declared in the class and defined out-of-line in `debug.hpp` -- the #1388 shape.
- **`assign()`'s "assign TBD" stub** prints with `std::printf` instead of `std::cout`:
  same destination, same text, no stream header.

## Two structural finds, and both cost the core its zero

**1. `numeric_limits.hpp` included the `areal.hpp` UMBRELLA.** Since the umbrella includes
the core, that is a cycle -- `core -> numeric_limits -> umbrella -> manipulators +
iostream` -- that `#pragma once` merely masked. It left `core.hpp` measuring **82,614
lines with all five I/O-family headers, seven lines short of the umbrella itself**. It now
includes `areal_impl.hpp`, which is what it actually needs.

This is the #1427 trap from the other direction: there, `iostream.hpp` included
`manipulators.hpp` back; here a *core* header reached the umbrella. Worth checking for in
the remaining types.

**2. `native/integers.hpp`, which carries `<sstream>`, was unconditional but is used only
by the trace blocks.** It moved inside the guard -- along with `native/ieee754.hpp`,
because the trace calls `to_binary(float)`, which lives in the text half of ieee754 that
the core's `ieee754_core.hpp` deliberately omits. **Without that second include
`TRACE_CONVERSION=1` stopped compiling** -- a regression this PR introduced, caught by the
trace gate, and verified against `main` (which compiles it fine) before fixing.

## Also

- `to_string()` and `to_binary()` are free functions formatting through a stringstream ->
  `manipulators.hpp`.
- `operator<<`/`operator>>` keep their friend **declarations** in the class (`operator>>`
  reaches `_fraction` directly, so friendship is real here, unlike dbns); only the
  definitions move to `iostream.hpp`, which is what lets the core get by with `<iosfwd>`.
- `native/ieee754.hpp` -> `native/ieee754_core.hpp`; `AREAL_ENABLE_LITERALS` and
  `AREAL_THROW_ARITHMETIC_EXCEPTION` default in `areal_impl.hpp` (#1436);
  `manipulators.hpp` now names the headers it uses (#1389).

## Verification

- **Layer gate**: `core.hpp` alone, and `core.hpp` + `manipulators.hpp`, each compiled,
  linked and run under `-Wall -Wpedantic` on gcc 13.3 and clang. The second gate
  **instantiates every manipulator** rather than merely including the header -- the
  strengthening #1449 forced after `range()` reached review with an undetected dependency
  on `iostream.hpp`.
- **Trace gate**: `TRACE_CONVERSION=1` compiles and prints on both compilers. This is what
  caught find 2.
- **Self-containment**: core, manipulators, iostream, debug, numeric_limits, table and the
  umbrella each compile standalone with 0 errors on both compilers.
- **26 targets built and run green on both compilers**, including `edu_tables_areals` and
  the education example, which exercise the text layers.
- **Differential against `main`**: five configurations over their *entire* encoding space
  (`areal<8,2>`, `<12,4>`, `<16,5>`, `<16,8>` exhaustively; `<20,6>` over 65,536), each
  rendered by `to_binary` (both nibble settings), `to_string`, `to_hex` (three settings),
  `hex_print`, `pretty_print`, `color_print`, `info_print`, `operator<<` under four stream
  states, the native conversions and the predicates; plus a dynamic-range round trip, the
  relocated `constexprClassParameters()` report and the `assign()` stub.

  **72,780,424 bytes byte-identical** -- old vs new on gcc, old vs new on clang, and gcc
  vs clang.
- **Trace text**: unlike dbns, `main` can switch areal's tracing on as shipped, so the
  comparison is direct. **67,398 bytes over 129 conversions across two configurations,
  byte-identical on both compilers.**

## Found, not fixed

`components(areal)` does not compile: `blockbinary<v.fbits> f;` uses a non-static member
in a constant expression, and it calls a `decode()` overload that does not exist. It fails
**identically on `main`** -- nothing in the tree instantiates it. Same class as
`type_field(bfloat16)` in #1445; both want one issue about manipulator templates that no
test ever instantiates.

## Summary by CodeRabbit

- **New Features**
  - Added dedicated stream input/output support for areal values, including exact values, uncertainty intervals, and NaN.
  - Added string and binary formatting utilities, with optional nibble separators for binary output.
  - Added debugging output for compile-time configuration and representation details.

- **Refactor**
  - Separated arithmetic, formatting, stream I/O, and debugging functionality into focused headers while preserving existing areal capabilities.
  - Improved header self-containment and configuration handling.

### DIFF
--- a/include/sw/universal/number/areal/areal.hpp
+++ b/include/sw/universal/number/areal/areal.hpp
@@ -15,28 +15,27 @@
 
 ////////////////////////////////////////////////////////////////////////////////////////
 ///  BEHAVIORAL COMPILATION SWITCHES
-
-////////////////////////////////////////////////////////////////////////////////////////
-// enable/disable the ability to use literals in binary logic and arithmetic operators
-#if !defined(AREAL_ENABLE_LITERALS)
-// default is to enable them
-#define AREAL_ENABLE_LITERALS 1
-#endif
-
-////////////////////////////////////////////////////////////////////////////////////////
-// enable throwing specific exceptions for integer arithmetic errors
-// left to application to enable
-#if !defined(AREAL_THROW_ARITHMETIC_EXCEPTION)
-// default is to use std::cerr for signalling an error
-#define AREAL_THROW_ARITHMETIC_EXCEPTION 0
-#endif
+///
+/// AREAL_ENABLE_LITERALS and AREAL_THROW_ARITHMETIC_EXCEPTION now default in
+/// areal_impl.hpp, beside the code they govern, so that a translation unit which includes
+/// core.hpp directly gets the same defaults this umbrella used to supply (#1334, #1436).
+/// Defining either before this header still wins, as before.
+///
+/// TRACE_CONVERSION, which switches on the conversion tracing, has always defaulted in
+/// areal_impl.hpp; its <iostream> include now sits inside that guard too.
 
 ////////////////////////////////////////////////////////////////////////////////////////
 /// INCLUDE FILES that make up the library
-#include <universal/number/areal/exceptions.hpp>
-#include <universal/number/areal/areal_impl.hpp>
-#include <universal/number/areal/numeric_limits.hpp>
+// layer 1: the arithmetic core (#1334). Include core.hpp directly in a translation
+// unit that only computes -- it pulls no <iostream>/<sstream>/<iomanip>.
+#include <universal/number/areal/core.hpp>
+
+// layer 2a: the string producers -- to_string, to_binary, to_hex, pretty_print, color_print
 #include <universal/number/areal/manipulators.hpp>
+// layer 2b: the <iostream> half -- operator<< / operator>>
+#include <universal/number/areal/iostream.hpp>
+// layer 3: introspection -- constexprClassParameters
+#include <universal/number/areal/debug.hpp>
 
 ///////////////////////////////////////////////////////////////////////////////////////
 /// math functions
--- a/include/sw/universal/number/areal/areal_impl.hpp
+++ b/include/sw/universal/number/areal/areal_impl.hpp
@@ -1,7 +1,21 @@
 #pragma once
-#include <iostream>   // std::cout/cerr used below (#1334: include what you use)
+
+// Behavioural switches default HERE, beside the code they govern, rather than in the
+// areal.hpp umbrella: including core.hpp directly would otherwise leave them undefined,
+// #if would evaluate them as 0, and the switch would silently flip on that path -- the
+// trap #1390 hit with POSIT_ENABLE_LITERALS (#1334, #1436).
+#if !defined(AREAL_ENABLE_LITERALS)
+#define AREAL_ENABLE_LITERALS 1
+#endif
+#if !defined(AREAL_THROW_ARITHMETIC_EXCEPTION)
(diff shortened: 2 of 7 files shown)


Ravenwater:
CodeRabbit's `operator>>` finding is **correct, and worse than it describes -- but pre-existing and identical on `main`.** Filed as #1454.

`std::cin >> someAreal` has never worked. It is a **link error**, not a compile error, which is why nothing has caught it:

```
undefined reference to `std::istream& sw::universal::operator>><16u, 5u, unsigned short>(
    std::istream&, sw::universal::areal<16u, 5u, unsigned short>&)'
```

I reproduced that byte-for-byte against a worktree of `main` and against this branch -- the same undefined symbol, the same mangled signature, both sides.

Two independent defects, as CodeRabbit says:

1. The friend **declaration** takes `areal&`; the **definition** takes `const areal&`. They are different functions, so extraction into a mutable areal selects the declared one, which is never defined.
2. The definition would not compile anyway -- `istr >> v._fraction` and areal has no `_fraction`; its storage is `bt _block[nrBlocks]`.

This PR relocates the definition from `areal_impl.hpp` to `iostream.hpp` and changes nothing inside it; the friend declaration stays in the class. Both are verbatim from `main`.

Not fixed here because it is design work, not a refactor. `operator<<` emits `[d]` for an exact areal and `(d, dnext)` for an uncertain one -- the ubit is the point of the type -- so a faithful `operator>>` has to decide whether it parses that interval syntax back or only accepts a decimal and clears the ubit, and those are observably different. There is no `parse(std::string, areal&)` to build on (unlike bfloat16/dfloat/hfloat), and `assign()` is itself a stub that prints `"assign TBD"`. #1454 lays out the sequence: pick the grammar, add `parse()`, make `assign()` call it, align the signatures, add the regression test that would have caught this.

Related: #1453, on `components(areal)` and `type_field(bfloat16)` -- manipulator templates that no test instantiates. Same root cause, and the reason all three of these are surfacing now is the strengthened layer gate rather than the test suite.

PR #1511, opened 2026-09-15
fix(areal): an exact zero sum is -0 only when both zeros are -0

### LINKED ISSUE #1507: areal: +0 - +0 and +0 + -0 give -0, where IEEE 754 gives +0
## Summary

areal `+0 - +0` and `+0 + -0` give `-0`. IEEE 754 gives `+0` for both in every rounding mode except toward negative: an exact zero sum of operands with opposite signs is `+0`.

```cpp
areal<16, 5, std::uint16_t> z(0.0);
z - z;      // [-0], IEEE: +0
z + (-z);   // [-0], IEEE: +0
```

Measured on `main` (7daa877a), and identical across every configuration tried. Over every pair of exact values, it is the only zero-result mismatch: `x - x` for nonzero `x` already gives `+0`.

`operator-=` is `*this += -rhs`, so this is the `+0 + -0` case, reached through negation. The sign rule is the one #1491 applied to dfloat. `x + 0 = x`, and two zeros sum to `-0` only when both are `-0`.

Whether areal should follow IEEE here is a decision for the maintainer; areal already departs from IEEE in its NaN equality (see `operator==` in `areal_impl.hpp`). If it should, the fix is in the zero handling of `operator+=`.

Found while fixing #1503.

Ravenwater:
## Summary

`+0 + -0` and `+0 - +0` gave `-0` (#1507). IEEE-754 gives `+0` for both in every rounding mode except toward negative, and areal truncates toward zero.

`operator+=` returned `rhs` whenever `*this` was zero, so `+0 + -0` copied the `-0`. `operator-=` is `*this += -rhs`, so `+0 - +0` reached the same case. Two exact zeros now sum to `-0` only when both are `-0`.

| | before | after | IEEE |
|---|---|---|---|
| `+0 + -0` | `-0` | `+0` | `+0` |
| `+0 - +0` | `-0` | `+0` | `+0` |
| `-0 + -0`, `-0 - +0` | `-0` | `-0` | `-0` |
| `-0 + +0`, `-0 - -0`, `+0 - -0`, `x - x` | `+0` | `+0` | `+0` |

The ubit interval `(0, minpos)` is unaffected, because `iszero()` does not match it.

## Test

`saturation.cpp`'s exact-reference sweep of `+ - *` no longer accepts either sign on an exact zero, and its operands now include `-0` as well as `+0`. The check is sign-exact: the double reference carries IEEE's zero signs, and areal's `==` compares encodings.
- Without this fix, the sweep fails exactly the two cases above in every configuration.
- With it, everything passes at every level on gcc and clang.
- The areal suite passes 26/26, and the areal consumers print identical output to main.

## Behaviour change

`+0 + -0` and `+0 - +0` now give `+0` instead of `-0`.

Closes #1507

https://claude.ai/code/session_01NazYbxaii2XXVzQeQKMDTs


Ravenwater:
@coderabbitai review

PR #1530, opened 2026-09-16
test(rational): enumerate the raw component encodings, and let base8, base10 and base16 gate

### LINKED ISSUE #1526: rational tests: make ValidateAssignment radix-aware, and let the base8/base10/base16 copies gate
## Summary

`ValidateAssignment` in the four `static/rational/*/conversion/assignment.cpp` copies is written for the base2 rational and does not fit the other bases. Raised by CodeRabbit on #1524, which fixed only the immediate hang.

**1. The bound is the wrong space.** The loops run to `NR_ENCODINGS = 1ull << nbits`, and `nbits` in the base8, base10 and base16 specializations is a compatibility alias for `ndigits`:

```cpp
static constexpr unsigned ndigits = _ndigits;
static constexpr unsigned nbits   = _ndigits;  // alias to make generic code easier to write
using Component = blockdigit<ndigits, 16>;
```

So for `rational<8, base16>` the sweep walks 256 numerator and 256 denominator **values** through `set(long long, long long)`, which normalizes them. It never enumerates `blockdigit<8,16>` encodings, of which there are 16^8 per component. The `static_assert(nbits <= 20, ...)` guards the alias, not the real state space.

**2. Those three copies never gate.** `static/rational/{octal,decimal,hexadecimal}/conversion/assignment.cpp` have `#define MANUAL_TESTING 1` committed, so they print failures and return `EXIT_SUCCESS`, and their REGRESSION_LEVEL_1 blocks are dead code. (The binary copy was switched to gating in #1524.)

**3. Two of them swept the wrong type.** The octal and hexadecimal copies called `ValidateAssignment<rb8>` and `<rb16>`, and `rb8`/`rb16` are the **base2** aliases (`rational.hpp:62`). #1524 pointed them at their own base, because once base2 conversion became exact the 2^32 `rb16` sweep no longer exited early and ran for hours. Per-base aliases already exist and should be used: `ro8`/`ro16` (base8, line 69), `rd8`/`rd16` (base10, line 75), `rh8`/`rh16` (base16, line 81).

## Suggested direction

- Give `ValidateAssignment` a radix-aware bound: derive the component space from the specialization's radix and `ndigits` (`radix^ndigits`), and enumerate raw `Component` encodings rather than values passed through `set()`. That needs a raw setter on the base8/base10/base16 specializations if one does not exist.
- Choose a digit count whose Cartesian product is feasible: `base16` with 3 digits is 4096 encodings per component, about 16.7M pairs; `base8` with 4 digits is the same order. 8 digits is not.
- Use the per-base aliases, not the base2 ones.

## Before those copies can gate

The base8, base10 and base16 specializations convert from a native float by truncating to an integer:

```cpp
rational& operator=(double rhs) { if (std::isfinite(rhs)) { n = static_cast<long long>(rhs); } else { n = 0; } d = 1; return *this; }
```

So 0.75 becomes 0/1. Turning `MANUAL_TESTING` off for those three files will make them fail until those conversions are implemented, e.g. with the best-rational-approximation search #1523 added for base2 (`convert_ieee754` in the base2 specialization). Worth deciding whether that is this issue or a separate one.

Related: #1523 / #1524, #1525.

### LINKED ISSUE #1528: blockdigit: a trial multiple that overflows the digit vector wraps around, so division answers too large a quotient digit
## Summary

`blockdigit::div_mod()` finds each quotient digit by adding the shifted divisor to a running multiple until the multiple passes the accumulator. `operator+=` is modular -- it drops the carry out of the top digit -- so a multiple that overflows the digit vector wraps around to a small value, still compares under the accumulator, and the loop keeps going. The quotient digit it settles on is then too large.

```cpp
for (unsigned m = 1; m < radix; ++m) {
    blockdigit candidate(multiple);
    candidate += subtractand;                              // modular: wraps at radix^ndigits
    if (candidate.compare_magnitude(accumulator) <= 0) {   // a wrapped value looks like it fits
        multiple = candidate;
        k = static_cast<DigitType>(m);
    }
    else break;
}
```

## Reproduction

```cpp
#include <universal/internal/blockdigit/blockdigit.hpp>
using namespace sw::universal;
blockdigit<2, 8> a(50), b(3);
std::cout << (long long)(a / b) << " rem " << (long long)(a % b) << '\n';   // 38 rem 0, expected 16 rem 2
blockdigit<2, 8> c(63), d(63);
std::cout << (long long)(c / d) << '\n';                                    // 7, expected 1
```

For `50 / 3` in a `blockdigit<2, 8>`: the subtractand is `3 << 1 == 24`, and the trial multiples run 24, 48, then 72, which wraps to 8. 8 is under 50, so the loop accepts it and goes on to 32, ending at k = 4 instead of k = 2, and the quotient comes out 38 remainder 0.

The correct multiple never overflows -- `k * subtractand <= accumulator <= max` -- so only the probing past the correct k can wrap. A positive subtractand can fail to grow the multiple only by wrapping, which is the signal to stop.

## Scope

Any division whose divisor, shifted up to the dividend's magnitude, has a multiple that exceeds `radix^ndigits - 1`. That is the common case once the operands use most of the digit vector, and it is invisible in a wide type with small values, which is why the existing hand-picked cases in `internal/blockdigit/arithmetic/arithmetic.cpp` pass: they run in a `blockdigit<8, radix>` with operands of at most 1234.

Consumers that divide: the base8, base10 and base16 `rational` specializations (`normalize()` reduces by the gcd, and the gcd loop is built out of `%`), `blockdecimal`, `blockoctal`, `blockhexadecimal`, and `positional`.

It surfaced while making `ValidateAssignment` radix-aware for #1526: enumerating the raw `blockdigit<ndigits, radix>` encodings puts operands at the top of the range, where the wrap happens, and every base failed the round trip -- `rational<2, base16>` 4312 times out of 39k normalized pairs. It is not visible in the value sweeps of #1524 / #1527, which only reach 2^ndigits.

## Fix

Stop the trial loop when the candidate stops growing, and add an exhaustive division regression over narrow configurations (every dividend and divisor a `blockdigit<2, radix>` can hold, against native integer division), where the wrap actually fires.

Fixed as a prerequisite in the #1526 part 2 PR, since the three digit-based rational assignment suites cannot gate until division is right.

Ravenwater:
## Summary
Part 2 of #1526, rebased onto `main` now that #1527 has merged (it gave base8, base10 and base16 a real float conversion). Replaces #1529, which GitHub closed when its base branch was deleted on that merge. `ValidateAssignment` was written for the base2 rational and had been copy-pasted into four files; it now lives in one verification header and enumerates whatever encoding space the type actually has.

- **base2** keeps the bit-pattern sweep over `2^nbits` values.
- **base8 / base10 / base16** enumerate the raw `blockdigit<ndigits, radix>` encodings: `radix^ndigits` per component, both signs, written with `setdigit()` and handed to the constructor that does not normalize. The same normalized-only filter applies (a zero denominator is the NaN encoding, a pair with a common factor comes back in lowest terms, negative zero comes back positive).

That distinction is the point of the issue: `nbits` in the digit specializations is a compatibility alias for `ndigits`, so the old sweep walked `2^ndigits` numerator and denominator **values** through `set()`. For a `rational<8, base16>` that was 256 of the 16^8 encodings a component holds, and none of the ones near the top of the range.

`MANUAL_TESTING 1` is gone from the octal, decimal and hexadecimal copies, so all four suites gate. The per-base aliases replace the base2 `rb8` / `rb16` the octal and hexadecimal copies referenced, and each suite reports under its own name instead of three of them sharing "rational float assignment validation".

## What enumerating the real space found

Every base failed immediately, and not by a rounding hair:

| type | encoding | `double(a)` | came back as |
|---|---|---|---|
| `rational<2, base8>` | 50/3 | 16.666666666666668 | **38/1** |
| `rational<2, base10>` | 65/6 | 10.833333333333334 | **21/2** |
| `rational<2, base16>` | 241/3 | 80.333333333333329 | **251/1** |

`rational<2, base16>` failed 4312 of its ~39k normalized pairs. The conversion was innocent: `best_rational_from_native` returned p = 50, q = 3 and `to_digit_component` encoded them correctly, but `set(50, 3)` -- which is just `normalize()` -- also answered 38/1.

The root cause is in `blockdigit::div_mod()` and is filed as #1528. It finds each quotient digit by adding the shifted divisor to a running multiple until the multiple passes the accumulator, and `operator+=` is modular: the multiple wraps at `radix^ndigits` and the wrapped value still compares under the accumulator, so the loop keeps going and takes too large a digit. For `50 / 3` in a `blockdigit<2,8>` the trial multiples run 24, 48, then 72, which wraps to 8; the loop continues to 32 and settles on k = 4 instead of k = 2. `63 / 63` answered 7.

The fix is one line plus its reasoning: the correct multiple never overflows (`k * subtractand` is at most the accumulator), so only the probing past the correct k can wrap, and a positive subtractand can only fail to grow the multiple by wrapping. `internal/blockdigit/arithmetic/arithmetic.cpp` gets an exhaustive division regression -- every dividend and divisor a `blockdigit<2, radix>` holds, against native integer division -- because the existing hand-picked cases run in a `blockdigit<8, radix>` with operands of at most 1234, where the digit vector is nowhere near full and the wrap cannot happen. Against the pre-fix header that regression fails at `-63 / -63 = 7`; with the fix all three radices pass.

This is a separate commit and could be split into its own PR, but the three digit-based suites cannot gate without it.

## Sweep sizes

Measured on this machine, all reporting zero failures:

| configuration | encodings per component | time | level |
|---|---|---|---|
| `rational<2, base8>` | 64 | 0.13 s | 1 |
| `rational<2, base10>` | 100 | 0.39 s | 1 |
| `rational<2, base16>` | 256 | 2.34 s | 1 |
| `rational<3, base8>` | 512 | 10.9 s | 2 |
| `rational<3, base10>` | 1000 | 43 s | 2 |
| `rational<4, base8>` | 4096 | 823 s | 4 |
| `rational<3, base16>` | 4096 | 776 s | 4 |

The 4096-encoding configurations are the ~16.7M pairs the issue suggested; each conversion runs a continued fraction in a 256-bit `blockbinary`, so they belong in level 4, not in the sanity tier. At level 1, which is what CI runs, the three new suites cost about 3 seconds together.

## Changes
- `include/sw/universal/verification/rational_test_suite.hpp` -- new: `ValidateAssignment` dispatching to the bit-pattern sweep or the raw-encoding sweep.
- `static/rational/{binary,octal,decimal,hexadecimal}/conversion/assignment.cpp` -- use the shared header, drop the four local copies, gate, per-base aliases, per-level digit counts, distinct suite names.
- `include/sw/universal/internal/blockdigit/blockdigit.hpp` -- stop the trial multiple before it wraps (#1528).
- `internal/blockdigit/arithmetic/arithmetic.cpp` -- exhaustive division regression.

## Test Results
| suite | gcc | clang |
|---|---|---|
| full regression, 401 tests | 401/401 PASS | 401/401 PASS |
| `rational_assignment`, `ro_assignment`, `rd_assignment`, `rh_assignment` | PASS | PASS |
| `blockdigit_arithmetic` | PASS | PASS |

- Levels 2 and 4 of the new sweeps were run out of band (the table above): zero failures in every configuration, including both 16.7M-pair sweeps.
- Cross-architecture under QEMU: aarch64 (binary128), ppc64le (double-double) and riscv64 (`LONG_DOUBLE_SUPPORT` 0) all pass the four assignment suites and `blockdigit_arithmetic`.
- `tools/scripts/check-ascii.sh`: clean.

## Test plan
- [ ] Promote to ready

Resolves #1526
Resolves #1528

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved division handling to prevent wrapped intermediate values from being accepted as valid quotient results.

- **Tests**
  - Added comprehensive validation for rational-number assignments through floating-point conversion, including binary, octal, decimal, and hexadecimal formats.
  - Expanded exhaustive division testing across multiple numeric bases and digit sizes.
  - Refined regression coverage and test naming for clearer rational-format validation.


Ravenwater:
Follow-up: the `encoding_span()` overflow CodeRabbit flagged is fixed in #1531 (a configuration past the sweep limit now answers one over it, so the caller's static_assert rejects it instead of accepting a wrapped 0 and sweeping nothing).

